Issue #33, opened 2021-12-14
Not working with MagicMouse, Macbook Pro, Big Sur


rachid-square:
I installed the application and the mouse acceleration and sensitivity settings have no effect. I tried restarting my computer, and toggling the accessibility settings of the app.

lujjjh:
Thank you for reporting this. Would you like to paste the relevant logs so that I can dig this further?

1. Open Console.app
2. Select the device and click Start streaming
3. Search CursorManager and choose Category

<img width="887" alt="image" src="https://user-images.githubusercontent.com/3000535/146114159-00146c2a-76e9-479f-9c1e-f8df569c05be.png">

You may see some logs (about every 5 seconds) describing which devices' acceleration and sensitivity are updated and which devices' are skipped. I'm guessing that Magic Mouse is skipped.

rachid-square:
I am also not seeing anything in the console when I follow those steps.

lujjjh:
Oops...I forgot to point out that View → Include Debug Messages should be selected in Console.app.

rachid-square:
For the record I found that option under "Action" -> "Include Debug Messages"

Here is the message:

> Magic Mouse 2 is skipped as it might be a trackpad

lujjjh:
> For the record I found that option under "Action" -> "Include Debug Messages"
> 
> Here is the message:
> 
> > Magic Mouse 2 is skipped as it might be a trackpad

Thank you for your reply. I attempted to fix this in #36. I don't have a Magic Mouse, though, so I can't test it. Please let me know if [this fix](https://github.com/lujjjh/LinearMouse/actions/runs/1586574138) works.

acefer:
I tried the new build wit the fix and the console app show this messages.
<img width="1387" alt="Captura de pantalla 2021-12-16 a las 12 32 59" src="https://user-images.githubusercontent.com/60737788/146364527-336be2cb-7d4f-4d99-8b18-94a5ee14d817.png">

lujjjh:
Magic Mouse appears to be quite different from a regular mouse. Perhaps this is the reason why SteerMouse and USB Overdrive do not support Magic Mouse (and Magic Trackpad). I'm going to mark this issue help wanted because it's probably not easy to fix.

However, I'm still curious about the structure of Magic Mouse in I/O registry. It might be useful to run

```sh
ioreg -n "AppleMultitouchDevice" # or maybe it's not an AppleMultitouchDevice
```

[...]

PR #36, opened 2021-12-16
Support Magic Mouse

lujjjh:
Try fixing #33.


lujjjh:
The fix doesn't work.

PR #75, opened 2022-02-09
Add German translation

aerulion:
I'm not sure if this was the intended way of contributing, or if I should have created an issue first, but here's my take on the German translation.


lujjjh:
Thank you very much!

PR #149, opened 2022-06-15
Support per-device settings

lujjjh:


lujjjh:
## Device switching

image

PR #254, opened 2022-09-20
Support matching parent process and process group

ViRb3:
I wanted to disable mouse acceleration on Minecraft, while keeping it enabled for all other apps. Since the game is executed by the Java runtime, the current solution is to add Java's bundle id to a LinearMouse rule. The problem with this is that the rule will cover all Java apps, not just Minecraft. You can't distinguish between Java-based apps because they all use the same Java runtime and binary.

My solution is to use the fact that typically, Java apps are started through a launcher, and this launcher will be set as the parent process. For example, if I use the Minecraft launcher PolyMC, the process graph will look like this:
```
- org.polymc.PolyMC
   L com.azul.zulu.java
```
This PR will use the C API for sysctl to look up the top-most parent process of every foremost window, and use that to apply the LinearMouse rule. This works perfectly for my use case in PolyMC on a MacBook Pro M1 Max (macOS 12.6).

There is one issue — with this change, it is no longer possible to target the leaf process (Java), or any parent that is not the top process. To fix this, we can enumerate the parent only if no rule matches the current process. Basically, we check the parent until we reach a process that matches a rule. I am not a Swift developer, so I didn't attempt to do this, but it should be pretty trivial. I am hoping you could help me with this.

Let me know if you have any feedback or suggestions. Thanks!


lujjjh:
Thank you very much for contributing!

I think this is a smart solution to match Java programs like Minecraft. There are some minor flaws, like if you quit the launcher after the game starts, you won't be able to get the parent process, but this should be an edge case.

It is worth noting that there is similar logic in EventTransformer, but it is in a hot path, so there may be performance issues if we enumerate the parent processes in it:

https://github.com/linearmouse/linearmouse/blob/d098e9f3ae2e044f922846b4a168611a179cf5f1/LinearMouse/EventTransformer/EventTransformer.swift#L19-L22

I'd prefer to provide a configuration that explicitly matches the parent process (or the process group). This would circumvent the issue you mentioned, as well as the potential performance issues.

For example:

```json
{
  "if": {
    "app": {
      "bundleID": "com.azul.zulu.java",
      "parent": {
        "bundleID": "org.polymc.PolyMC",
        "parent": {
          ...
        }
      }
    }
  }
}
```

Or simpler, but not nestable:

```json
{
  "if": {
    "app": "com.azul.zulu.java",
    "parentApp": "org.polymc.PolyMC",
    "groupApp": "org.polymc.PolyMC"
  }
}
```

What do you think?

Of course, there are some alternative ways I can think of. For example, matching the command-line arguments or the working directory of the process, but this is obviously more complicated.

ViRb3:
> I'd prefer to provide a configuration that explicitly matches the parent process (or the process group). This would circumvent the issue you mentioned, as well as the potential performance issues.

I like this approach! Both options would suit my use case, so I don't have a preference, but maybe the first one is better since it's more flexible?

Would you be able to implement this functionality given your knowledge of the codebase? Hopefully it won't be much effort. Otherwise I can look into doing it myself, but it will take me a while...

lujjjh:
> maybe the first one is better since it's more flexible 

It seems that the first one so flexible that I couldn't find an actual use case. The latter is easier to use and has a constant performance. So, I finally chose the latter.

ViRb3:
Looks good to me. It's easier to understand your approach as well. Thanks a lot for implementing this!

Issue #403, opened 2023-03-31
Feature Request: Add ability to remap mouse buttons for macOS gestures (Quick Look, Zoom)


C1ares:
### OS

macOS 13.2.1 (Build 22D68)

### LinearMouse

v0.8.0

### Is your feature request related to a problem?

I am a macOS user who frequently uses gestures like Quick Look and Zoom. I would like to use LinearMouse to remap some of my mouse buttons to trigger these gestures easily. 

### Describe the solution you'd like

I propose adding the ability to remap mouse buttons to (trackpad) macOS gestures. E.g. mb4 click to quick look, hold and scroll for zoom (like pinch in and out on trackpad)

### Describe alternatives you've considered

An alternative solution would be to use third-party software like BetterTouchTool or Mac Mouse Fix to achieve this functionality. However, integrating this feature directly into LinearMouse would provide a more streamlined experience and reduce the amount of potentially conflicting background apps

hotsno:
Hi @lujjjh, do you mind sharing the status on this issue?

I recently discovered LinearMouse and have really been enjoying it. It almost completely replaces the 3 tools I was using prior (SteelSeries ExactMouse, Discrete Scroll, and Mac Mouse Fix). However, I am really missing the Cmd + Scrollwheel to trackpad zoom feature that [Mac Mouse Fix](https://github.com/noah-nuebling/mac-mouse-fix) provides. Trackpad zoom is oftentimes superior to normal zoom on web browsers imo. Their project is open source so that might be helpful.

lujjjh:
Hey @hotsno! I'm glad you like LinearMouse. I did try to support trackpad zoom, but it wasn't as straightforward to implement as other gestures, and I personally prefer normal zoom in most cases. So, I temporarily put it on hold at the time.

However, I can understand that you prefer trackpad zoom more. I would likely to start by simply add support for remapping modifier keys + scroll wheel to trackpad zoom, rather than getting into more complex scenarios like remapping the hold of mb5 + scroll wheel.

hotsno:
Sounds good, thank you very much for your work!

hotsno:
Hey @lujjjh, I just installed the beta version of LinearMouse, and tried out the Pinch Zoom option with Command modifier key! It zooms great, but it continues to zoom even after I let go of the Command key. Just wanted to report in case you weren't aware.

PR #602, opened 2023-09-16
Scrolling modifiers: Add support for simulating pinch zoom gestures

lujjjh:
Issue: #403
Comment: https://github.com/linearmouse/linearmouse/issues/403#issuecomment-1718377346

### DIFF
--- a/LinearMouse/EventTransformer/ModifierActionsTransformer.swift
+++ b/LinearMouse/EventTransformer/ModifierActionsTransformer.swift
@@ -2,19 +2,32 @@
 // Copyright (c) 2021-2023 LinearMouse
 
 import Foundation
+import GestureKit
 import KeyKit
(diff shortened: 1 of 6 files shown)


mknev:
Would it be possible to add the ability to use a mouse button as the modifier to enable pinch to zoom? This would be great as it could then be done completely one-handed. Many thanks!

TheFutureIsVoluntary:
> Would it be possible to add the ability to use a mouse button as the modifier to enable pinch to zoom? This would be great as it could then be done completely one-handed. Many thanks!

I second this suggestion. Well, on a feature to allow the mouse button as modifier to enable anything - I'd like left button + scroll up to be able to trigger "Mission Control". And right button + scroll up -> move to left display / scroll down -> move to right display.

PR #651, opened 2023-11-22
Add support for additional configuration paths

### LINKED ISSUE #644: I want to locate configuration file to `~/Library/Application Support`
### OS

macOS 14.1.1

### LinearMouse

v0.9.5

_No response_

I want to locate configuration file to `~/Library/Application Support/linearmouse` instead of `~/.config/linearmouse`.

#### Why?
In macOS, most applications locate configuration files to `~/Library/Application Support/`, but LinearMouse not.
I want to be consistent with the location of my configuration files.

#### How?
If `~/Library/Application Support/lienearmouse/linearmouse.json` exists, please use it, otherwise use current location. To avoid breaking the experience for existing users.

_No response_

### Additional context

_No response_

lujjjh:
The configuration file will be located in the order below:

1. ~/Library/Application Support/linearmouse/linearmouse.json
2. ~/.config/linearmouse/linearmouse.json

This commit enables the use of additional configuration locations, including the standard ~/Library/Application Support path, to provide users with more flexibility.

The configuration file can now be located as follows:

1. ~/Library/Application Support/linearmouse/linearmouse.json
2. ~/.config/linearmouse/linearmouse.json (The original location)

To maintain backward compatibility and prevent any disruptions, new configurations will still be created in location 2.

Closes #644.

### DIFF
--- a/LinearMouse/EventTransformer/SwitchPrimaryAndSecondaryButtonsTransformer.swift
+++ b/LinearMouse/EventTransformer/SwitchPrimaryAndSecondaryButtonsTransformer.swift
@@ -4,11 +4,13 @@
 import Foundation
 import os.log
 
-enum SwitchPrimaryAndSecondaryButtonsTransformer {
+class SwitchPrimaryAndSecondaryButtonsTransformer {
     static let log = OSLog(
         subsystem: Bundle.main.bundleIdentifier!,
         category: "SwitchPrimaryAndSecondaryButtonsTransformer"
     )
+
+    init() {}
 }
 
 extension SwitchPrimaryAndSecondaryButtonsTransformer: EventTransformer {


ultratiem:
Apple guidelines dictate that any config files or files in general needed by an app install to ~/Library/Application Support. That is the standard and ubiquitous location for all application files on macOS.

macOS does not use hidden directories nor should it. Nor does it install to a users home folder. That is a Linux standard.

This commit leads to user confusion as now there are 2 config files. Moreover, the secondary location is not ideal as macOS users should have to look for hidden directories.

This commit should leave the correct location (~/Library/Application Support), and remove the .config folder entirely. There is almost no gain for a user to move config files or store them in another location. All this will do is add a layer of confusion and takes the app further away from one designed exclusively for macOS.

with respect.

lujjjh:
Hi @ultratiem! Thanks for your feedback.

To clarify, this pull request marks the beginning of the migration of the configuration from `~/.config` to `~/Library/Application Support`. However, to avoid introducing breaking changes before the next major update, it is necessary to keep the original configuration path.

PR #756, opened 2024-07-01
update zh-Hans

Tinyu-Zhao:
Although the direct translation of "Switch primary and secondary buttons" in English does mean "交换主要和次要按键", this will make most people confused, what are the primary and secondary buttons, if translated as "交换鼠标左右按键" will make it easier to understand

SCR-20240701-qbay


lujjjh:
Thanks!

PR #834, opened 2024-10-27
Switch per-app settings from focused app to app under cursor

lujjjh:
- Previously: Settings based on focused app (using `eventTargetUnixProcessID`)
- Now: Settings based on app under mouse cursor

This change makes per-app settings more intuitive. Users can now move
the cursor over an app and immediately use its specific settings (e.g.,
custom scrolling distances) without changing focus.


pypykim:
This merge helped me solve a big problem

PR #841, opened 2024-11-11
added non latin chars to keyboard shortcuts

### LINKED ISSUE #840: Keyboard Shortcuts with chars like Ö, Ä, Ü, ß as not recorded and ignored.
### OS

macOS 14.6.1 (23G93)

### LinearMouse

v0.10.0

### Describe the bug

I wanted to use the Back Button of my Mouse to navigate back inside the Finder. As I'm on a German UI Setting, the Shortcut for that is `CMD` + `Ö` which will not be recorded by LinearMouse but plain ignored.

My wild guess is, that LinearMouse does not cover the keys or something other fails.

### To reproduce

1. Open the Keyboard Settings in LinearMouse
2. Add a Mouse Action
3. Press back for example
4. Choose key command from the Action Select
5. Press a non-Latin character like Ö, Ä, Ü, ß and see that its not recorded.

### Expected behavior

Should work with any key 

### Anything else?

_No response_

appinteractive:
Trying to fix #840 

The shortcut of ö, ä ü, etc. is now correctly displayed.

### DIFF
--- a/Modules/KeyKit/Sources/KeyKit/Key.swift
+++ b/Modules/KeyKit/Sources/KeyKit/Key.swift
@@ -92,7 +92,7 @@ public enum Key: String, Codable {
     case å
     // Polish keyboards:
     case ą
-    case ć  
+    case ć
     case ę
     case ł
     case ń


appinteractive:
For some reason, the action that I do setup does disappear after some seconds. Does the Debug kills the settings from a local installation?

lujjjh:
> For some reason, the action that I do setup does disappear after some seconds. Does the Debug kills the settings from a local installation?

This might be due to the configuration not being forward-compatible. The new configuration may not be parsed by the version you have installed locally, causing it to revert to an empty configuration.

appinteractive:
> Thank you for the pull request! I expected the shortcut configuration to be independent of keyboard layout. I'll look into it later.

It seemed to be limited by the configuration of the key enum as it might be mapped to only those values from what I could grasp. I'm not that familiar with Swift to be honest.

lujjjh:
The changes look reasonable to me. Could you address the lint error?

appinteractive:
Done, linter should be happy now

PR #864, opened 2024-12-28
added support for kensington slimblade top buttons

### LINKED ISSUE #853: Kensington Slimblade Trackball Top Left and Right buttons aren't recognized (unable to assign custom actions)
### OS

mac 15.1

### LinearMouse

v0.10.0

### Describe the bug

Unable to assign actions top two mouse buttons on kensington slimblade trackball "mouse" (they aren't detected as input).

### To reproduce

press assign action to mouse buttons and press top right or left button and nothing happens.

### Expected behavior

for it to detect the inputs

### Anything else?

_No response_

fraschm1998:
Closes #853 

Not sure if another device category should be added for trackballs instead of adding the device to productsToApplySideButtonFixes as they're not really "side buttons"

### DIFF
--- a/LinearMouse/Device/Device.swift
+++ b/LinearMouse/Device/Device.swift
@@ -21,7 +21,8 @@ class Device {
 
     private static let productsToApplySideButtonFixes: Set<Product> = [
         .init(vendorID: 0x2717, productID: 0x5014), // Mi Silent Mouse
-        .init(vendorID: 0x248A, productID: 0x8266) // Delux M729DB mouse
+        .init(vendorID: 0x248A, productID: 0x8266), // Delux M729DB mouse
+        .init(vendorID: 0x047d, productID: 0x2041)  // Kensington Slimblade
     ]
 
     private weak var manager: DeviceManager?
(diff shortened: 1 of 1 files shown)


lujjjh:
Thanks for your contribution!

> Not sure if another device category should be added for trackballs instead of adding the device to productsToApplySideButtonFixes as they're not really "side buttons"

I agree—it might be more appropriate to create a separate `Set<(VendorID, ProductID)>` to store the devices requiring this fix. This would help avoid using hardcoded magic numbers directly in the code and improve clarity and maintainability.

fraschm1998:
@lujjjh should be good now

fraschm1998:
@lujjjh so far I just have as I don't have any other trackball devices to try. And I'm not sure if I check for productName contains "trackball" will it mess up devices that have only 2 buttons and 2 scroll reverse buttons.

image 

```
       if let vendorID = device.vendorID,
            let productID = device.productID,
            Self.trackballsRequiringReportMonitoring.contains(
                .init(vendorID: vendorID, productID: productID))
        {
            return .trackball
        }
```

I could do something like:

```
        if isTrackball {
            return .trackball
        }

  private var isTrackball: Bool {
        // First check our known device list
        if let vendorID = device.vendorID,
           let productID = device.productID,
           Self.trackballsRequiringReportMonitoring.contains(
               .init(vendorID: vendorID, productID: productID))
        {
            return true
        }
        
        // Then check device characteristics
        if device.confirmsTo(kHIDPage_GenericDesktop, kHIDUsage_GD_Mouse) {
            // Check if it explicitly identifies as a trackball in name
            if self.productName?.lowercased().contains("trackball") == true {
                return true
            }
        }
        
        return false
    }
```

but then again I don't know if it will mess up the trackball "mice"

lujjjh:
Thank you for your effort, and Happy New Year!

After giving it some thought, I wonder if parsing the device's report descriptor might be worth exploring. It could potentially handle extra button recognition for a broader range of devices more effectively. Your work here has provided a great starting point, and I might look into this approach over the weekend or when I have some time.

PR #929, opened 2025-05-23
Mouse move pointer events can be forwarded to the scroll wheel

patrick99e99:
I have two trackballs. I wanted one to behave as normal for mouse movement and clicking, and the other to act as a dedicated scroll device.  I could not find any way to do this with any existing software, so I forked this project and added a "Pointer redirects to scroll" option under the pointer configuration settings.

I have extensively tested this functionality and it does exactly what I needed, so I thought I would open a pull request for it to see if there is interest in having this added to the project.

### DIFF
--- a/LinearMouse/EventTap/GlobalEventTap.swift
+++ b/LinearMouse/EventTap/GlobalEventTap.swift
@@ -16,7 +16,7 @@ class GlobalEventTap {
 
     init() {}
 
-    private func callback(event: CGEvent) -> CGEvent? {
+    private func callback(_ event: CGEvent) -> CGEvent? {
         let mouseEventView = MouseEventView(event)
         let eventTransformer = EventTransformerManager.shared.get(
             withCGEvent: event,
@@ -43,14 +43,13 @@ class GlobalEventTap {
             return
         }
 
+        var eventTypes: [CGEventType] = EventType.all
+        if SchemeState.shared.schemes.contains(where: { $0.pointer.redirectsToScroll ?? false }) {
+            eventTypes.append(EventType.mouseMoved)
+        }
+
         do {
-            observationToken = try EventTap.observe([.mouseMoved, .scrollWheel,
-                                                     .leftMouseDown, .leftMouseUp, .leftMouseDragged,
-                                                     .rightMouseDown, .rightMouseUp, .rightMouseDragged,
-                                                     .otherMouseDown, .otherMouseUp, .otherMouseDragged,
-                                                     .keyDown, .keyUp, .flagsChanged]) { [weak self] _, event in
-                self?.callback(event: event)
-            }
+            observationToken = try EventTap.observe(eventTypes) { [weak self] in self?.callback($1) }
         } catch {
             NSAlert(error: error).runModal()
         }
--- a/LinearMouse/EventTransformer/PointerRedirectsToScrollTransformer.swift
+++ b/LinearMouse/EventTransformer/PointerRedirectsToScrollTransformer.swift
@@ -37,7 +37,6 @@ class PointerRedirectsToScrollTransformer: EventTransformer {
     private func topLeftScreenCoordinates() -> CGPoint {
         let mouseLocation = NSEvent.mouseLocation
         guard let screen = NSScreen.main else { return mouseLocation }
-        let screenHeight = screen.frame.height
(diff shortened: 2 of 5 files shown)


Modules/PointerKit/Sources/PointerKit/PointerDevice.swift
```diff
@@ -190,6 +190,18 @@ public extension PointerDevice {
                                       forKey: pointerAccelerationType ?? kIOHIDMouseAccelerationTypeKey)
         }
     }
+
+    /**
+     Indicates whether pointer X/Y events should be redirected to scroll X/Y events.
+     */
+    var pointerRedirectsToScroll: Int? {
+        get {
+            client.getProperty("HIDRedirectsPointerToScroll")
```
lujjjh, line 199: There doesn't seem to be a `HIDRedirectsPointerToScroll` property. Maybe we can remove this property definition from `PointerDevice`.

LinearMouse/Device/Device.swift
```diff
@@ -152,6 +152,18 @@ extension Device {
         }
     }
 
+    var pointerRedirectsToScroll: Bool? {
```
lujjjh, line 155: And this property can be removed also.

LinearMouse/EventTap/GlobalEventTap.swift
```diff
@@ -44,7 +44,7 @@ class GlobalEventTap {
         }
 
         do {
-            observationToken = try EventTap.observe([.scrollWheel,
+            observationToken = try EventTap.observe([.mouseMoved, .scrollWheel,
```
lujjjh, line 47: I’m a bit concerned about performance—specifically CPU usage. Have you tested whether moving the mouse with this feature disabled causes any noticeable CPU spike?

LinearMouse/EventTransformer/PointerRedirectsToScrollTransformer.swift
```diff
@@ -0,0 +1,43 @@
...
+
+        if let scrollEvent = CGEvent(
+            scrollWheelEvent2Source: nil,
+            units: .pixel,
+            wheelCount: 2,
+            wheel1: Int32(scrollY),
+            wheel2: Int32(scrollX),
+            wheel3: 0
+        ) {
+            scrollEvent.post(tap: .cghidEventTap)
```
lujjjh, line 31: It would be ideal if this event could be propagated to the subsequent event transformers, allowing the scroll features to be applied as well.

Modules/PointerKit/Sources/PointerKit/PointerDevice.swift
patrick99e99: oh-- yes.. that was a misunderstanding on my part.  I have removed this.

LinearMouse/EventTap/GlobalEventTap.swift
patrick99e99: I did profile, and saw a few spikes here and there, but nothing that seemed to terrible...  In any event (pun intended), I changed this in a new commit so that the `.mouseMoved` event only gets added to the events list if a device has that redirects to scroll setting turned on.  (this required the global event tap to be stopped and restarted whenever that `scheme.pointer.redirectsToScroll` property changes)

patrick99e99:
> A nitpicky suggestion: this feature might be better placed as a separate section below the pointer speed/acceleration section. Logically, pointer speed and acceleration belong together, and this feature doesn’t really need to be affected by the "Revert to system defaults" option.

Well, there is the potential for someone with one device to turn this setting on, and then suddenly they will not be able to move their mouse because it will be frozen sending scroll events instead of pointer move events!  So it seems like we would want "Revert to system defaults" to apply to this so that it can be unset by pressing that `^+shift+cmd+Z` key combination.

lujjjh:
> Well, there is the potential for someone with one device to turn this setting on, and then suddenly they will not be able to move their mouse to turn it off because it would then be frozen [...]

Actually, that shouldn’t be an issue—right after you enable this setting, even if movement is locked you can still click the toggle with the left button to turn it off. You’re absolutely right, though, that we’ll need to think about how someone without a second mouse could disable it later. Thanks for your contribution—I’ll go ahead and merge this now, and we can revisit the UI placement afterward.

Issue #906, opened 2025-03-17
LM thinks two mice with the same model number are the same mouse


AdamRealE:
### OS

macOs Sequoia 15.3.1 (24D70)

### LinearMouse

v0.10.0

### Describe the bug

I use two external mice, both are Logitech M650 L.
One is right handed, one is left handed.
I only use one at a time, but when I switch I need to toggle: 

> Settings > Buttons > Switch primary and secondary buttons

because it thinks they are one mouse.

This is probably not a common problem.

### To reproduce

1. Two Logitech M650 L mice
2. Disconnect one
3. Connect the other
4. See that it is sharing the same settings as the previous device

### Expected behavior

Each mouse to have their own settings.

### Anything else?

[LinearMouse.log](https://github.com/user-attachments/files/19296189/LinearMouse.log)

- I enabled verbose logging,
- switched from the left to right handed mouse, 
- hit the toggle to switch mouse primary click, 
- switched back to the left handed mouse, 
- hit the toggle to switch the buttons back, 
- then exported the last 5 minutes.

PR #940, opened 2025-06-02
allow multiple identical devices to be configured

patrick99e99:
After my last pull request was merged in for the `pointerRedirectsToScroll` feature, I realized that if you have two identical devices (such as two trackballs that are the same model from the same manufacturer), the configuration object currently is shared for BOTH devices, so impossible to make on handle mouse events, and the other redirect to scroll.

The changes in this PR fix this behavior, so now if there are multiple identical devices then they will add a number so the items show up like "CST Laser Trackball #1" and "CST Laser Trackball #2" in the settings UI.

In order to ensure a unique configuration object was created for each duplicate device, I needed to include something in the `productName` property to differentiate the devices, so I used the device's `id` property for that.

Thanks.


lujjjh:
Thanks a lot for the contribution!

I'm not sure prepending the `id` in `productName` can reliably solve the issue you described. The `id` is not a stable value—it depends on the order in which devices are connected or discovered—so it’s not suitable for serialization into persistent storage like the configuration file.

When I originally designed this, I considered using `serialNumber` to distinguish identical devices. While not all devices have it (for example, my Logitech USB receiver), in my experience, devices connected via Bluetooth usually do provide a serial number.

Also, this change introduces a breaking change: since the `productName` would change, existing configuration schemes would stop working unless users manually update their configuration.

patrick99e99:
> I'm not sure prepending the `id` in `productName` can reliably solve the issue you described. The `id` is not a stable value—it depends on the order in which devices are connected or discovered—so it’s not suitable for serialization into persistent storage like the configuration file.

Yeah, I was just doing some testing and noticed that and made a change (force pushed) to basically just add the `#1` or `#2` at the end IF there are duplicates...

> Also, this change introduces a breaking change: since the `productName` would change, existing configuration schemes would stop working unless users manually update their configuration.

Well, what I just changed would now only break configurations IF people had multiple identical devices--  which might be unlikely?

lujjjh:
Thanks for your quick response.

I'm still not convinced `duplicateIdentifier` is stable. It relies on device connection order, which can lead to issues like reconnected devices being assigned new, duplicate identifiers (e.g., reconnected `#1` becoming `#2`). This means configurations could be randomly applied after restarts, which isn't ideal.

I agree that it makes sense to distinguish between devices with the same product name in the UI. However, our first step should probably be figuring out how to get a stable serial number / identifier for them. If you can shed some light on why the device isn't acquiring a serial number, we might be able to offer more targeted solutions.

patrick99e99:
> I'm still not convinced `duplicateIdentifier` is stable. It relies on device connection order, which can lead to issues like reconnected devices being assigned new, duplicate identifiers (e.g., reconnected `#1` becoming `#2`). This means configurations could be randomly applied after restarts, which isn't ideal.

Yeah, that's something I did not consider, so this PR does seem like the wrong approach...
 
> I agree that it makes sense to distinguish between devices with the same product name in the UI. However, our first step should probably be figuring out how to get a stable serial number / identifier for them. If you can shed some light on why the device isn't acquiring a serial number, we might be able to offer more targeted solutions.

Unfortunately, it looks like the CST Laser Trackball devices don't have serial numbers:
Screenshot 2025-06-02 at 10 47 12 AM

How about the commit I just pushed up now?  I added 'locationID' to devices and made the DeviceMatcher fallback to locationID if serialNumber is nil..  What do you think of that approach?

lujjjh:
Adding `locationId` matching sounds promising! I'll test this change when I have time, and thank you again for your contribution.

However, the logic around implementing `locationId` matching will be quite complex. This is mainly because:

- If we unconditionally add `locationId` matching: Every time users plug a device into a different USB port, the system would recognize it as a "new" device. This would force them to reconfigure all their devices, significantly impacting the user experience.
- If we determine whether to add `locationId` matching based on duplicate devices: While this approach sounds more reasonable, it could complicate future adjustments to existing configuration schemes or require us to re-evaluate how current configurations are applied.

Ideally, we'll need a more comprehensive scheme management user interface to effectively handle these scenarios.

I also noticed #906. This `locationId` approach might not fully address all the use cases there, especially when devices are connected via the same port (meaning their `locationId` is identical). In such scenarios, it would still be impossible to distinguish between them. A more robust, though labor-intensive, solution would be to adapt to each manufacturer's private protocol—like hidpp—to obtain individual device serial numbers.

PR #1011, opened 2025-09-22
Sidebar: Chagne to have Pointer item as first item

jrogby:
Most often you want to adjust your pointer settings, and not the scrolling settings.


lujjjh:
Thanks, that makes sense to me. However, you’ll also need to update the default navigation item here:

https://github.com/linearmouse/linearmouse/blob/171efc4075d2dac0594640331ed425b46b3fea04/LinearMouse/State/SettingsState.swift#L13

It might also make sense to persist the default navigation item according to the last selected item.

PR #1211, opened 2026-05-29
Allow Logitech gesture buttons to preserve click fallback

ehardesty:
## Summary

Allow Logitech gesture-button controls to behave like regular mouse buttons when no gesture is detected.

Previously, Logitech virtual/gesture controls were treated as fully handled as soon as the gesture transformer saw them. That prevented the Logitech fallback synthetic click from being posted, so a configured gesture button could trigger gestures but would not preserve normal click behavior.

This changes Logitech control handling from a boolean handled/not-handled result to a small enum so handlers can distinguish between:

- controls that are fully handled internally
- controls that are not handled
- gesture controls that are handled for gesture tracking but should still allow synthetic fallback click events

With this, a Logitech gesture button can:

- fall through as a click when released without crossing the gesture threshold
- trigger the configured gesture when dragged past the threshold
- suppress the click action after a gesture fires

## Testing

- `xcodebuild test -project LinearMouse.xcodeproj -scheme LinearMouse`
- `swiftformat --lint` on changed Swift files
- `swiftlint --strict` on changed non-legacy Swift files

### DIFF
--- a/LinearMouse/Device/VendorSpecific/Logitech/LogitechHIDPPDeviceMetadataProvider.swift
+++ b/LinearMouse/Device/VendorSpecific/Logitech/LogitechHIDPPDeviceMetadataProvider.swift
@@ -2133,11 +2133,11 @@ final class LogitechReprogrammableControlsMonitor {
                             modifierFlags: modifierFlags
                         )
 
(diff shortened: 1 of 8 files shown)


lujjjh:
Thanks!

PR #1212, opened 2026-05-30
Add Logitech hardware DPI support

ehardesty:
## Summary

Adds device-specific hardware DPI support for Logitech HID++ devices that expose the Adjustable DPI feature.

This change:
- detects adjustable DPI support for direct Logitech HID++ devices and receiver-routed devices
- reads current DPI and the supported DPI range from the selected device
- adds a Hardware DPI control to Pointer settings only when the selected device supports adjustable DPI
- applies and persists DPI per selected device, not globally or per app/display
- reapplies saved DPI when configuration changes or a matching device is connected
- extends configuration with `pointer.hardwareDPI`
- adds focused unit coverage for DPI list parsing, current DPI reads, DPI writes, sorting/filtering, and fallback behavior

## Device Testing

Manually tested with:
- Logitech MX Ergo S Plus
- Logitech MX Master 3S

## Behavior Notes

- Unsupported devices do not show the Hardware DPI UI.
- The DPI field is seeded from the current hardware DPI.
- Applying DPI writes the nearest supported value, reads back the current DPI, and only saves after read-back succeeds.
- Saved DPI is stored on the base device-specific scheme so it follows the selected device switcher rather than app/display context.
- HID++ DPI operations are serialized on a dedicated queue to avoid blocking the UI.

## Verification

- `xcodebuild -project LinearMouse.xcodeproj -scheme LinearMouse -quiet`
- `xcodebuild test -project LinearMouse.xcodeproj -scheme LinearMouse -only-testing:LinearMouseUnitTests/LogitechHIDPPDeviceDPIControllerTests -quiet`
- `swiftformat .`
- `swiftlint lint LinearMouse/Device/VendorSpecific/Logitech/LogitechHIDPPDeviceDPIController.swift`
- `swiftlint lint LinearMouse/UI/PointerSettings/PointerSettings.swift LinearMouse/UI/PointerSettings/PointerSettingsState.swift`


lujjjh:
Thanks for the PR!

I gave it a quick try, but the Hardware DPI control didn’t show up for my M720. I’m not sure whether this is because the implementation is still incomplete.

Maybe we’ll eventually need the concept of a logical device, so that `DeviceMatcher` can support matching a specific logical device behind a receiver.

I’ll take a closer look at the implementation when I have more time.

ehardesty:
Was the M720 using bluetooth or the unifying receiver? I was only able to test with bluetooth due to my devices using the bolt receiver.
I can iterate on this PR a bit. I'll also dig around to see if I can find a unifying receiver device.

i was planning to tackle bolt receiver support after this but I may consider adding that first, then the dpi support for a broader receiver base to work with.

ehardesty:
After investigating a bit, it looks like the M720 does not support adjusting hardware DPI via HID++
https://github.com/pwr-Solaar/Solaar/blob/master/docs/devices/M720%20Triathlon%20Multi-Device%20Mouse%20405E.txt

In contrast the MX Master 3 I tested with does advertise the adjustable dpi flag:
https://github.com/pwr-Solaar/Solaar/blob/master/docs/devices/MX%20Master%203S%20B034.txt

lujjjh:
It works perfectly on my ERGO M575. Thanks so much for your contribution!

PR #1261, opened 2026-07-03
fix: ensure CMD keyup always fires when non-Latin input causes unsupportedKey

### LINKED ISSUE #1078: Keyboard shortucts cmd+ via tilt buttons get stuck
### OS

macOS 15.5 (Build 24F74)

### LinearMouse

v0.10.3-beta.1

### Describe the bug

Quite regularly I noticed a weird behaviour related to linearmouse. The keyboard combination tied to tilt buttons on my mouse is cmd + [ and cmd + ] (to emulate back and forth) and the cmd get stuck. This results in unusable keyboard and system as a whole. App relaunch solves everything, but it's annoying. Could be a bug on beta, will roll back to 10.2 to doublecheck

### To reproduce

_No response_

### Expected behavior

_No response_

### Anything else?

_No response_

ssw112:
Fixes #1078

## Problem

When a keyboard shortcut action (e.g. `cmd+[`) is triggered via a mouse button and the active input method is non-Latin (e.g. Ukrainian), `postKeyLocked` throws `unsupportedKey` for the character key (`[`). Because `press()` uses `try` without error handling, the exception exits the keyDown loop early — `CMD` keydown has already fired but the keyUp loop never runs. `CMD` stays stuck in `CGEventFlags` until LinearMouse is relaunched.

## Fix

Wrap both the keyDown and keyUp loops in `press()` with individual `do/catch` blocks so the keyUp loop always executes regardless of any keyDown failures. Keys that failed to press down will silently no-op on keyUp (harmless).

Also adds `import os` for `os_log` error logging.

## Repro

1. Set a mouse button to keyboard shortcut `cmd+[`
2. Switch input method to a non-Latin language (e.g. Ukrainian)
3. Click the mouse button → CMD gets permanently stuck until app relaunch

### DIFF
--- a/Modules/KeyKit/Sources/KeyKit/KeySimulator.swift
+++ b/Modules/KeyKit/Sources/KeyKit/KeySimulator.swift
@@ -2,6 +2,7 @@
 // Copyright (c) 2021-2026 LinearMouse
 
 import AppKit
+import os
 
 public enum KeySimulatorError: Error {
     case unsupportedKey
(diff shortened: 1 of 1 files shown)


lujjjh:
Thanks!

PR #1298, opened 2026-07-23
Prevent stale process metadata after PID reuse

offyotto:
Process metadata and event transformers were cached only by PID. When macOS reuses a PID, LinearMouse can read the previous process metadata or return its transformer before matching the new process.

Resolve one process identity containing the PID and start time, then use it for metadata lookup, scheme matching, and transformer cache keys. Repeated events from the same process still reuse their cached values.

Test: DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcodebuild test -quiet -project LinearMouse.xcodeproj -scheme LinearMouse -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO (305 passed)

### DIFF
--- a/LinearMouse/EventTransformer/EventTransformerManager.swift
+++ b/LinearMouse/EventTransformer/EventTransformerManager.swift
@@ -19,7 +19,7 @@ class EventTransformerManager {
 
     struct CacheKey: Hashable {
         var deviceMatcher: DeviceMatcher?
-        var pid: pid_t?
+        var process: ProcessIdentity?
         var screen: String?
     }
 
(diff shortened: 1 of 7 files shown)


lujjjh:
Thanks for catching this! I think there may be one more cache affected by PID reuse: `EventTransformerManager` caches transformers by `(deviceMatcher, pid, screen)` and may return a stale transformer before `matchScheme` or the process metadata is checked.

Perhaps we could introduce a shared process identity containing the PID and start time, use it for both the metadata caches and `EventTransformerManager.CacheKey`, and resolve it once along the matching path. A regression test covering transformer caching across PID reuse would also be helpful.

offyotto:
Confirmed. `EventTransformerManager` checked `(deviceMatcher, pid, screen)` before `matchScheme`, so a reused PID could return the previous transformer. I updated the branch to use the same PID and start-time identity for both caches and added a regression test covering transformer reuse across PID recycling. Full suite: 305 passed.

Issue #708, opened 2024-03-04
Support for non standard HID Devices


bhedavivek:
### OS

MacOS Sonoma - 14.3.1 

### Is your feature request related to a problem?

Using a Elecom Huge Trackball mouse. Buttons 6-8 are not detected by LinearMouse. Further only mouse wheel  left tilt actions is detected as mouse wheel press. Need support for mouse wheel tilt detection. 

Alternate solutions like USBOverdrive, Steermouse are able detect them. Likely needs some device specific configuration in LinearMouse.

### Describe the solution you'd like

Buttons on trackball mouse with more than 5 standard buttons are detected by Linear Mouse during configuration.

aalpgiray:
on it

bhedavivek:
adds support for the wireless version of elecom huge trackball - https://github.com/linearmouse/linearmouse/pull/1320

PR #1302, opened 2026-07-25
Support ELECOM HUGE TrackBall Fn buttons

### LINKED ISSUE #1250: Fn buttons are not detected in Elecom Mouse Huge
### OS

macOS Versione 26.5.1 (Build 25F80)

### LinearMouse

v0.11.2

Fn buttons are not detected in Elecom Mouse Huge

Fn mouse buttons detected. 

_No response_

### Additional context

_No response_

aalpgiray:
## Summary

The ELECOM HUGE TrackBall has 8 buttons, but only 5 of them can be recorded or mapped in LinearMouse today. The remaining three (the Fn buttons) produce no event at all.

The cause is in the device's report descriptor. Its mouse collection declares only 5 buttons:

| Report ID | Contents |
| --- | --- |
| 1 | Button usages 1–5, X, Y, Wheel, AC Pan |
| 3 | System Power Down / Sleep / Wake |
| 4 | Vendor page `0xFEBC`, array of 47 usages |
| 5 | Consumer page, array of 573 usages |

Left, right, middle and the two thumb buttons take up all five declared buttons, so the Fn buttons cannot be buttons 6–8 in this layout. They are nevertheless reported in the *same* input report, in the three bits the descriptor declares as padding:

```
01 20 00 00 00 00 00 00     Fn button, byte 1 bit 5
01 00 00 00 00 00 00 00     released
01 40 00 00 00 00 00 00     Fn button, byte 1 bit 6
01 00 00 00 00 00 00 00     released
01 80 00 00 00 00 00 00     Fn button, byte 1 bit 7
01 00 00 00 00 00 00 00     released
```

macOS honours the descriptor and drops those bits, so no `otherMouseDown` is ever generated and there is nothing for the event tap to see. This is also why SteerMouse and USB Overdrive detect these buttons on a physically connected device — they read raw HID reports rather than CGEvents.

This PR adds an `InputReportHandler` that reads the three undeclared bits out of the raw input report and simulates button 5 to 7. It follows the same approach as the existing `KensingtonSlimbladeHandler`, which solves the same class of problem for the Slimblade's top buttons.

Two details worth noting for review:

- `alwaysNeedsReportObservation()` returns `true`. The device reports 5 buttons, so the existing `buttonCount == 3` heuristic in `Device` would never enable report observation for it.
- The handler only parses reports whose ID is `0x01`. The device also emits consumer and vendor-defined reports, whose second byte would otherwise be misread as button states.

## On making this more general

This deliberately stays device-scoped, because a general fix is a larger change than it first appears.

The mechanism itself — *buttons hidden in bits the descriptor declares as padding* — is not vendor specific, and it is exactly what `GenericSideButtonHandler` and `KensingtonSlimbladeHandler` already work around one device at a time. Handling it automatically for any device would mean parsing `kIOHIDReportDescriptorKey` to locate the button field and its trailing padding bits, then treating those bits as extra buttons. Because events are only emitted on a bit *toggle*, genuine padding that stays zero would produce nothing, so that approach is safer than it sounds.

The blocker is not the parsing but the observation cost: it would require enabling raw input report observation for every connected mouse, which the current architecture deliberately gates per device. That trade-off felt like it deserves its own PR and a maintainer's opinion rather than being bundled into a device fix, so I'm happy to follow up on #708 if you think the cost is acceptable.

There is also a reason to keep raw report handling as a *supplement* rather than a direction of travel. It only works for devices physically attached to the Mac running LinearMouse. Input forwarded from another Mac by Universal Control arrives as a virtual device — `V-HUGE TrackBall`, same VID/PID, in my case — and HID-level detection cannot reach it. That is reportedly why HID-only utilities do not work with continuity-forwarded devices at all, while LinearMouse's CGEvent-level transforms (reverse scrolling and so on) work fine on the receiving Mac. Whatever shape a general fix for #708 takes, the CGEvent pipeline should stay the primary path, with report parsing filling in only the buttons macOS discards.

In the meantime the product ID table in the handler makes other ELECOM trackballs a one-line addition once their layout is confirmed on hardware. I only own the HUGE, so I have not added models I cannot verify.

## Known limitation: forwarded input

The buttons this recovers exist only in the raw reports of a physically connected device, so they are only available on the Mac the trackball is attached to. In my testing, an Fn button mapped to middle click did not fire on a second Mac over Universal Control. I did not isolate whether Universal Control relays middle clicks at all, so I would not put that forward as a confirmed limitation of this change — but it should not be expected to work. Mapping the Fn buttons to keystrokes is the workaround, since keyboard input is forwarded.

## Testing

- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer xcodebuild test -project LinearMouse.xcodeproj -scheme LinearMouse -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO` — 315 passed. The 2 failures in `StatusItemBatteryIndicatorTests` are pre-existing and locale dependent (`"100 %"` vs `"100%"`, non-breaking space); they reproduce on an unmodified `main` in the same environment.
- `swiftformat --lint` and `swiftlint lint` clean on the changed files.
- 10 new unit tests covering each Fn bit, release on button up, simultaneous presses, unchanged state, the declared buttons 1–5 being left alone, and other report IDs being ignored.
- Verified on hardware with an ELECOM HUGE TrackBall (`0x056E:0x010C`, USB): all three Fn buttons are now recordable in Settings → Buttons as Button #5, #6 and #7, and mapped actions fire correctly.

Closes #1250

Refs #708 — this covers the button half of that report. The mouse wheel tilt detection also mentioned there is not addressed; the device reports tilt as AC Pan in report 1, which is a separate concern.

### DIFF
--- a/LinearMouse/Device/InputReportHandler.swift
+++ b/LinearMouse/Device/InputReportHandler.swift
@@ -59,7 +59,8 @@ extension InputReportHandler {
 enum InputReportHandlerRegistry {
     static let handlers: [InputReportHandler] = [
         GenericSideButtonHandler(),
-        KensingtonSlimbladeHandler()
+        KensingtonSlimbladeHandler(),
+        ElecomTrackballHandler()
     ]
 
     static func handlers(for vendorID: Int, productID: Int) -> [InputReportHandler] {
--- a/LinearMouse/Device/InputReportHandlers/ElecomTrackballHandler.swift
+++ b/LinearMouse/Device/InputReportHandlers/ElecomTrackballHandler.swift
@@ -1,0 +1,83 @@
+// MIT License
+// Copyright (c) 2021-2026 LinearMouse
+
+import Foundation
+
+/// Handles the extra Fn buttons of ELECOM trackballs.
+///
+/// The HID report descriptor of these devices declares only 5 buttons in its mouse
+/// collection, which are taken up by left, right, middle and the two thumb buttons.
+/// The remaining Fn buttons are still reported in the same input report, but in the
+/// three bits that the descriptor declares as padding, so macOS discards them and no
+/// `otherMouseDown` event is ever generated.
+///
+/// This handler reads those undeclared bits out of the raw input report and simulates
+/// the corresponding button events.
+///
+/// Other ELECOM trackballs share this report layout, so they can be supported by adding
+/// their product IDs below once the layout has been confirmed on the device.
+///
+/// Supported devices:
+/// - ELECOM HUGE TrackBall (0x056E:0x010C)
+struct ElecomTrackballHandler: InputReportHandler {
+    private struct Product: Hashable {
+        let vendorID: Int
+        let productID: Int
+    }
+
+    private static let supportedProducts: Set<Product> = [
+        .init(vendorID: 0x056E, productID: 0x010C) // ELECOM HUGE TrackBall
+    ]
+
+    /// Report format: | Button 0 (1 bit) | ... | Button 4 (1 bit) | Fn buttons (3 bits) |
+    ///
+    /// Buttons 0 to 4 are declared in the report descriptor and handled by macOS, so only
+    /// the upper three bits are of interest here.
+    private static let fnButtonsMask: UInt8 = 0xE0
+
+    /// Bit 5 to 7 are mapped to button 5 to 7, the first button numbers not already used by
+    /// the buttons macOS recognizes.
+    private static let fnButtonBits = 5 ... 7
+
+    private let emit: MouseButtonEmitter
+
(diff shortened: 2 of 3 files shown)


aalpgiray:
This is a topic I have no experience I though I just give it a try with ai and test. I have tested and seems to work. Let me know if there are things you want me to do to merge this. Great app by the way!

lujjjh:
Thanks for contributing! I don’t see any issues with the changes. Looks good to merge.

aalpgiray:
How does release cycles work? When this will be released?

lujjjh:
@aalpgiray Beta releases don’t follow a fixed schedule, but I’m planning to publish one with this change shortly.

Issue #1303, opened 2026-07-26
Scroll acceleration and inertia remain active when set to 0


rafajrichard:
### OS

macOS 26.5.2 (25F84)

### LinearMouse

v0.11.3

### Describe the bug

Scroll acceleration and inertia still affect smooth scrolling when both settings are set to 0. Faster wheel input produces a greater scrolling distance for the same number of wheel ticks, and scrolling continues after wheel input stops.

### To reproduce

1. Enable smooth scrolling.
2. Set Scroll response to Immediate.
3. Set Scroll acceleration and Inertia to 0.
4. Slowly scroll a fixed number of wheel ticks and note the distance travelled.
5. Repeat with the same number of wheel ticks, but scroll them quickly.
6. Observe that the resulting distance differs and that scrolling continues after wheel input stops.

### Expected behavior

With acceleration and inertia set to 0, each wheel tick should produce a consistent scrolling distance regardless of input speed, and scrolling should stop when wheel input stops. Smooth scrolling should only animate the discrete wheel movement rather than introduce additional acceleration or momentum comparable to emulating trackpad scrolling.

### Anything else?

Browser-native smooth scrolling in Chromium and Firefox demonstrates the intended type of behavior: wheel input is animated smoothly, but the scrolling distance remains tied to the wheel input and does not attempt to emulate trackpad momentum.

Solaris-star:
I traced this to the smoothed engine rather than settings persistence: `acceleration = 0` still uses the wheel-rate projection and the preset `inputExponent`, while `inertia = 0` still clamps decay to at least `0.72`, so both reported effects are expected from the current implementation. Before changing the physics contract, could a maintainer confirm whether slider value 0 is intended to mean **disabled** (no rate-dependent distance and no synthetic momentum), or only the current UI labels **Flat** / **Short**? The former requires an explicit zero-value engine path and regression tests for equal tick-distance plus no momentum; the latter would call for UI/help text clarification instead.

YuriNachos:
Looking into this. The rate-dependent amplification (`pow(normalizedMagnitude, profile.inputExponent)`, `SmoothedScrollingEngine.swift` ~L85) and the `momentumDecay` clamp to `0.72...decayCeiling` (~L178) both fire regardless of the slider, so `acceleration=0` / `inertia=0` don't actually disable them.

[...]

PR #1304, opened 2026-07-27
fix(scroll): respect zero scroll acceleration and inertia in smooth scrolling

amir-rezaei:
## Description
This PR resolves issue #1303 where scroll acceleration and inertia remain active even when set to 0.

## Details
- Disables the acceleration curve (magnitude calculation) in `desiredVelocity` when acceleration is explicitly set to 0, ensuring consistent scroll distance relative to input speed.
- Decays momentum immediately to 0 when inertia is set to 0, stopping scroll animation instantly when wheel input stops.


lujjjh:

Issue #1301, opened 2026-07-25
Can't use shift-scroll to horizontal scroll


JazzyStrat:
### OS

macOS 15.6.1 (Build 24G90)

### LinearMouse

v0.11.4-beta.3

### Describe the bug

holding shift while scrolling on the vertical wheel doesn't do anything with Logi mx master 3S. Shift modifier set to default action.

lujjjh:
Which scrolling mode are you using? For Smoothed or By Pixels, Shift must be set to “Alter orientation.” You can find this note by clicking the information icon. “Default” follows the system behavior.

JazzyStrat:
thank you very much, I missed the specific workout under the information icon

PR #1305, opened 2026-07-27
fix(modifier): default Shift key to alterOrientation for horizontal scroll compatibility

amir-rezaei:
## Description
This PR resolves issue #1301 by defaulting the Shift modifier to `alterOrientation` when no explicit action is defined (the `auto` / `nil` state).

## Details
- On macOS, holding Shift while scrolling natively changes vertical scroll into horizontal scroll.
- When LinearMouse handles scroll events (especially continuous/smoothed ones), the native macOS translation does not fire, causing Shift-scroll to stop working for many mice (like the MX Master 3S).
- By mapping Shift to `alterOrientation` as its default action, we preserve the native macOS Shift-scroll behavior out-of-the-box without requiring users to manually change configurations in the UI.


lujjjh:

Issue #1237, opened 2026-06-09
Auto Scroll Preserve Native Middle-Click Option Buggy


j3nnyryu:
### OS

macOS Tahoe 15.7.7

### LinearMouse

v0.7.5

### Describe the bug

<img width="847" height="612" alt="Image" src="https://github.com/user-attachments/assets/789177ed-bd59-4209-a67c-606a8b835654" />
are my settings
The middle click preserve native feature doesn't work unless I first drag the icon first, for links it works without dragging but for images and other icons such as tabs I need to drag or else it will bring up the auto scroll wheel.
****
<img width="1710" height="1107" alt="Image" src="https://github.com/user-attachments/assets/6f2a34a0-8d7d-482b-8059-fee21effb155" />

### To reproduce

1. Go to Youtube
2. Middle Click on a video icon
3. Doesn't open video on new tab like it's supposed to
4. UNLESS you drag the video icon first around with left click hold, then if you middle click the video icon again, it will work as expected for all icons that have been "touched".

This isn't only on Youtube, this issue persists among all websites and sometimes on the tab bar itself, like if you middle click the tab it should close out of it, but SOMETIMES it brings up the scroll wheel, I haven't been able to get it to perfectly reproduce like the Youtube though maybe because I already drag my tabs around.

### Expected behavior

Clicking on a video icon will open the video on a new tab.

PR #1306, opened 2026-07-27
fix(autoscroll): treat AXTab accessibility elements as pressable to preserve native middle-click

amir-rezaei:
## Description
This PR resolves part of issue #1237 by adding `AXTab` to the pressable and excluded accessibility roles in `AutoScrollAccessibilityActivationClassifier.swift`.

## Details
- Previously, `AXTab` was missing from `pressableRoles` and `excludedRoles`. As a result, middle-clicking browser tabs (e.g. in Safari, Chrome, Firefox) often triggered the AutoScroll wheel overlay instead of closing or opening tabs natively.
- Adding `AXTab` ensures tab UI elements are classified as pressable and preserve native middle-click handling.


lujjjh:

Issue #1210, opened 2026-05-29
Linear Mouse incorrectly sends spacebar keypresses when using modifier+scroll zoom, causing ⌘Space to trigger and launch Alfred/Spotlight.


4gotham:
### OS

macOS 26.4.1    

### LinearMouse

0.11.2

### Describe the bug

Linear Mouse incorrectly sends (CMD+spacebar) keypresses when using modifier+scroll zoom. This causes alfred app to launch while zooming. (My hotkey for alfred is CMD Space)

Evidence:
Confirmed via macOS Keyboard Viewer: CMD+spacebar visually highlights on occasion during CMD+scroll.  
It does not happen if you only scroll one time - happens with multiple scrolls....deeper zoom.

Impact:
Makes modifier+scroll zoom unusable due to constant accidental app launches.

### To reproduce

Steps to Reproduce:

Enable zoom with modifier key (Command or Option) in Linear Mouse

Open macOS Keyboard Viewer to monitor key presses

Hold modifier key (CMD) and scroll mouse wheel multiple times to zoom

### Expected behavior

Expected Behavior:
Only modifier key and scroll events should be sent.  AKA it should zoom in on the web page.

Actual Behavior:
It's occasionally sending CMD+SPACE (which annoyingly launches alfred) but it will still zoom.

lujjjh:
Thanks for the report. I tried to reproduce this locally, but haven’t been able to trigger the unexpected Space keypress so far.

Could you let me know which keyboard layout you’re using when this happens?

youtextme:
For Mac launcher buy/switch: a low-risk path is dual-hotkey — keep Spotlight or Alfred on ⌘Space, map free Raycast to ⌥Space for a week, and only pay for Alfred Powerpack or Raycast Pro after the free tier's clipboard/file actions actually stick in daily use.

https://raycast.com/?via=girish
Affiliate — I earn a commission if you upgrade to Pro.

PR #1308, opened 2026-07-27
fix(zoom): reset KeySimulator after simulated zoom keypresses to prevent modifier state leaks

amir-rezaei:
## Description
This PR resolves issue #1210 by calling `Self.keySimulator.reset()` after synthesizing zoom keypresses in `ModifierActionsTransformer.swift`.

## Details
- Previously, repeated scroll events during modifier+scroll zoom accumulated synthetic keypress state in `KeySimulator`.
- Calling `Self.keySimulator.reset()` ensures key state is cleared after each zoom event, preventing accidental modifier/spacebar leakage.


lujjjh:

PR #1334, opened 2026-08-05
Add support for a secondary gesture button and additional gesture actions

DrMatrix85:
This PR adds support for configuring a second independent gesture button alongside the existing gesture trigger.

The change updates ButtonMappingTransformer to support multiple GestureButtonTransformer instances while preserving the existing click-vs-gesture arbitration behavior.

It also adds new gesture actions:

Previous Track
Next Track
Maximize Window
Minimize Window

The secondary gesture profile has its own trigger, threshold, and directional action assignments.

Tested hardware

Tested with a Logitech MX Master 4.

Configuration used during testing:

Primary gesture trigger: standard mouse button
Secondary gesture trigger: Logitech HID++ Control 0x00C3
0x00C3 also mapped independently to Play/Pause on button release

Verified that:

both gesture buttons operate independently
normal short-click mappings continue to work on a gesture trigger
Previous/Next Track actions work with system media playback
Maximize resizes the focused window to the usable screen area
Minimize sends the focused window to the Dock
existing unit tests pass
SwiftFormat lint passes

Local make reaches a successful archive, but final export cannot complete because the local machine does not have the project’s Developer ID Application signing certificate.

### DIFF
--- a/LinearMouse/EventTransformer/ButtonMappingTransformer.swift
+++ b/LinearMouse/EventTransformer/ButtonMappingTransformer.swift
@@ -65,7 +65,7 @@ final class ButtonMappingTransformer: EventTransformer, DeferredEventTransformer
     private let monotonicClock: MonotonicClock
     private let fallbackEventSink: (CGEvent) -> Void
     private let swapsPrimaryAndSecondaryButtons: Bool
-    private let gestureTransformer: GestureButtonTransformer?
+    private let gestureTransformers: [GestureButtonTransformer]
     private let policy: ButtonMappingPolicy
 
     private var timer: TimerToken?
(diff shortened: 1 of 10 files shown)


lujjjh:
Thanks for working on this! I implemented support for multiple gesture-capable buttons in #1326 by extending Button Mappings with swipe triggers. Since this gesture is initiated by a button, it should now be possible to configure it through Button Mappings without introducing a separate secondary Gesture Button profile.

The additional actions in this PR (i.e. Previous/Next Track and Maximize/Minimize Window) would still be useful additions to the available Button Mapping actions. You’re very welcome to submit those separately in this or another PR.

DrMatrix85:
Thanks for the clarification! I didn't realize that #1326 generalized swipe triggers through Button Mappings. That makes sense and is a cleaner solution than adding a secondary Gesture Button profile.

I’ll move the Previous/Next Track and Maximize/Minimize functionality into the Button Mapping action system and submit those separately in a focused PR.

DrMatrix85:
Closing pull request- Processing requested changes to separate PR.

PR #1336, opened 2026-08-06
Restore LaunchAtLogin migration

### LINKED ISSUE #1328: Start at login silently breaks when macOS loses the registration; the app never re-registers (and #445's migration call was accidentally removed)
### OS

macOS 26.2 (25C56)

### LinearMouse

0.11.3

### Describe the bug

LinearMouse stopped launching at login on my machine (it had been working before). Rather than just re-toggle the setting, I dug into why — and I think the findings explain the recurring "start at login stopped working" reports (#441, #902) that previously went stale without a root cause.

**Machine evidence:** macOS's background-item database had *no record of LinearMouse at all* — `sfltool dumpbtm` returns zero matches for `com.lujjjh.LinearMouse` (no `SMAppService.mainApp` record, no legacy helper record, for any user), while other apps' login-item records are present. The app itself runs fine when launched manually, is installed in `/Applications`, and is not quarantined.

Reading the source, two things stand out:

**1. "Start at login" state is a live mirror of `SMAppService.mainApp.status`, so a lost registration is silent and permanent.**

LinearMouse uses the now-archived `sindresorhus/LaunchAtLogin` v5.0.0, whose `isEnabled` getter simply returns `SMAppService.mainApp.status == .enabled`. The user's intent is never persisted anywhere. macOS is known to occasionally drop background-item registrations (e.g. when the BTM database is reset/rebuilt, which can happen across OS updates). When that happens:

- LinearMouse never launches at login again;
- the "Start at login" toggle silently reads as off;
- nothing ever re-registers, because the app has no memory that the user enabled it.

(`register()` failures are also only visible in os.log, so an enable that fails looks identical to one that succeeded until the next reboot.)

**2. The macOS 13 migration call added in #445 was later removed, apparently accidentally.**

#445 added `LaunchAtLogin.migrateIfNeeded()` to `AppDelegate.applicationDidFinishLaunching`. Commit 5437d887 ("Fix window title visibility built on new Xcode", #835 — an unrelated change) removed that line; the removal shipped in v0.10.1 and the call hasn't existed since (the `import LaunchAtLogin` in `AppDelegate.swift` is now unused). Users coming from a legacy pre-macOS 13 registration are never migrated, and the `unregisterLegacy()` cleanup never runs.

### To reproduce

The registration loss itself is hard to trigger deterministically (it appears tied to BTM database resets during some macOS upgrades), but the resulting broken state is easy to verify on an affected machine:

```sh
sfltool dumpbtm | grep -ci linearmouse   # → 0, despite Start at login having been enabled
```

### Expected behavior

If the user enabled "Start at login", LinearMouse should notice on launch that the registration is missing and re-register (surfacing `SMAppService.Status.requiresApproval` if macOS wants confirmation in System Settings → Login Items).

### Anything else?

**Proposed fix — I'd be happy to send a PR if this direction is welcome:**

1. Persist the user's launch-at-login intent as a `Defaults` key (the project already uses Defaults).
2. On app launch, if intent is enabled but `SMAppService.mainApp.status != .enabled`, call `register()` again, and surface `.requiresApproval` to the user instead of failing silently.
3. Restore the `migrateIfNeeded()` call (or equivalent logic) removed in 5437d887.
4. Longer term, consider replacing the archived LaunchAtLogin package with a small internal `SMAppService`-based manager, keeping a legacy path only for macOS 11–12 (per the 11.0 deployment target).

**Workaround for anyone hitting this today** — a user LaunchAgent (keep the in-app "Start at login" toggle off so the two mechanisms don't race):

```xml
<!-- ~/Library/LaunchAgents/com.lujjjh.LinearMouse.plist -->
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>AssociatedBundleIdentifiers</key>
	<string>com.lujjjh.LinearMouse</string>
	<key>Label</key>
	<string>com.lujjjh.LinearMouse</string>
	<key>LimitLoadToSessionType</key>
	<string>Aqua</string>
	<key>ProcessType</key>
	<string>Interactive</string>
	<key>Program</key>
	<string>/Applications/LinearMouse.app/Contents/MacOS/LinearMouse</string>
	<key>RunAtLoad</key>
	<true/>
</dict>
</plist>
```

```sh
launchctl bootstrap gui/$(id -u) ~/Library/LaunchAgents/com.lujjjh.LinearMouse.plist
```

---

**Full disclosure:** this investigation and write-up are a human + Claude (Anthropic's AI assistant) collaboration. The evidence comes from a real affected machine and the code/commit references were checked against this repository, but I apologize in advance if any of the analysis misses the mark or if AI-assisted reports aren't welcome here — happy to amend or close if so. LinearMouse has been genuinely useful to me for years; this is just an attempt to give something back after the workaround fixed my setup.

YuriNachos:
## Summary
Commit 5437d887 (PR #835) accidentally deleted `AppDelegate`'s `override init()` that ran `LaunchAtLogin.migrateIfNeeded()`, so the macOS 13 legacy → `SMAppService` login-item migration stopped running and `import LaunchAtLogin` became a dead import.

This restores the migration on the launch path.

## Verification
`make configure clean lint test XCODEBUILD_ARGS="CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= DEVELOPMENT_TEAM="`
- `swiftformat --lint .` clean; `swiftlint .` 0 serious (96 pre-existing warnings, none in changed files).
- Full suite green: **531 tests, 0 failures** (8 new in `LinearMouseUnitTests/AppDelegate/LoginItemManagerTests`).

Closes #1328

### DIFF
--- a/LinearMouse/AppDelegate.swift
+++ b/LinearMouse/AppDelegate.swift
@@ -17,6 +17,14 @@ class AppDelegate: NSObject, NSApplicationDelegate {
     private var sessionActive = true
     private var sleeping = false
 
+    /// Runs the one-time legacy -> SMAppService login-item migration on launch.
+    ///
+    /// It's a no-op below macOS 13 and after the first successful run. This call
+    /// was lost in 5437d88 and is restored here (issue #1328).
+    override init() {
+        LaunchAtLogin.migrateIfNeeded()
+    }
+
     func applicationDidFinishLaunching(_: Notification) {
         guard ProcessEnvironment.isRunningApp else {
             return
--- a/LinearMouse/UI/StatusItem.swift
+++ b/LinearMouse/UI/StatusItem.swift
@@ -119,8 +119,11 @@ class StatusItem: NSObject, NSMenuDelegate {
         startAtLoginItem.state = LaunchAtLogin.isEnabled ? .on : .off
         LaunchAtLogin.publisher
             .receive(on: RunLoop.main)
-            .sink { [weak self] value in
-                self?.startAtLoginItem.state = value ? .on : .off
+            .sink { [weak self] _ in
+                // The publisher forwards the *requested* value even when the
+                // registration lands at `.requiresApproval` or fails, so re-read
+                // the actual `SMAppService` status to keep the checkmark truthful.
+                self?.startAtLoginItem.state = LaunchAtLogin.isEnabled ? .on : .off
             }
             .store(in: &subscriptions)
 


lujjjh:
Thanks for investigating this and for the thorough tests. Restoring `LaunchAtLogin.migrateIfNeeded()` is definitely useful, but that method only performs the one-time migration from the legacy login helper to `SMAppService`; it does not restore a modern registration that macOS has subsequently lost.

For the latter case, I’m not sure we should persist a separate intent and automatically re-register. Once the system status becomes `notRegistered`, we cannot reliably distinguish between macOS losing the registration and the user disabling it in System Settings. Automatically restoring it could therefore override an explicit user choice.

I think the simpler and safer approach would be to:

- Restore the `migrateIfNeeded()` call directly.
- Continue treating the actual system status as the single source of truth.
- Let the user enable “Start at login” again if the registration is lost.
- Surface `.requiresApproval` to the user when appropriate.

This would also avoid introducing the new Defaults key and the additional recovery state. One detail to watch is the menu item: `LaunchAtLogin.publisher` publishes the requested value even if registration fails, so it may not reflect the actual system status in that case.

Thanks again for working on this!

lujjjh:
Thanks!

PR #1337, opened 2026-08-06
Honor acceleration=0 and inertia=0 in SmoothedScrollingEngine

### LINKED ISSUE #1303: Scroll acceleration and inertia remain active when set to 0

YuriNachos:
## Summary
Honor the lower bound of the "Scroll acceleration" and "Inertia" sliders in `SmoothedScrollingEngine` so that the extreme "off" positions behave as users expect:

- **acceleration = 0** → bypass the rate-dependent `pow(normalizedMagnitude, inputExponent)` curve (and the acceleration gain, which is already `1` here) so each wheel tick travels a constant distance — velocity becomes strictly proportional to input.
- **inertia = 0** → `momentumDecay == 0`, so scrolling stops the moment the wheel stops, with no synthetic momentum phase.

Both changes sit behind a clean `== Scheme.Scrolling.Smoothed.<range>.lowerBound` short-circuit. The non-zero (legacy) code path is byte-for-byte unchanged, so the behavior is trivially reversible.

Root cause: `pow(normalizedMagnitude, profile.inputExponent)` amplified distance regardless of the acceleration slider, and `momentumDecay` was clamped to `0.72 ... decayCeiling`, leaving residual velocity even at inertia = 0.

## Verification
`SmoothedScrollingEngine.AxisTuning` was widened from `private` to `internal` so its velocity/decay contracts can be unit-tested via `@testable`. 8 focused tests were added:

- Linearity at acceleration=0 across 4 presets × 2 speeds (`v(100) == v(25) * 4`, `v(50) == v(25) * 2`).
- **Boundary just above lowerBound** (acceleration=0.001, inertia=0.001) still routes through the legacy curve — pins the `== lowerBound` contract so it can't silently widen to `<=`.
- Legacy curve still active at acceleration=1.2, pinned to golden value `3455.117`.
- `momentumDecay == 0` at inertia=0 (three `dt` values); end-to-end: no `momentumBegan` ever emitted.
- Legacy carry preserved at default inertia 0.65 (golden `pow(0.89, 0.5)`) and at the `0.25` clamp boundary.
- Mid-range regression checks for both acceleration and inertia.

Gate green: SwiftFormat + SwiftLint clean (0 serious; 95 pre-existing warnings in unrelated files); `SmoothedScrollingEngineTests` 27/0; full suite **531 tests, 0 failures**.

## Related issues
Closes #1303.

### DIFF
--- a/LinearMouse/EventTransformer/SmoothedScrollingEngine.swift
+++ b/LinearMouse/EventTransformer/SmoothedScrollingEngine.swift
@@ -41,7 +41,10 @@ final class SmoothedScrollingEngine {
         case smoothed(AxisTuning)
     }
 
-    private struct AxisTuning {
+    /// Internal so the velocity/decay contracts can be exercised directly via
+    /// @testable unit tests (see SmoothedScrollingEngineTests). It stays nested
+    /// under the engine and therefore out of the module's public API surface.
+    struct AxisTuning {
         private static let legacyUpperBound = 3.0
 
         let configuration: Scheme.Scrolling.Smoothed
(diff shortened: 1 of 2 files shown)


lujjjh:
Thanks! I was initially concerned that inertia 0 might make smooth scrolling stop too abruptly, but 0 does reasonably imply that inertia is fully disabled. I think this is reasonable to merge and evaluate based on user feedback.

Issue #1271, opened 2026-07-08
Middle click won't close Chrome tab when AutoScroll is enabled


Waley-Z:
### OS

macOS 14.8.4 and 26.4.1

### LinearMouse

v0.11.3-beta.6

### Describe the bug

This happend after Chrome is upgraded to 150.0.7871.100. #1263 does not seem to fix it for Chrome browser.

PR #1340, opened 2026-08-07
Fix Auto Scroll activation delay in Finder and add opt-in override for pressable elements

### LINKED ISSUE #1339: Auto Scroll does not work when middle-click dragging over clickable items/links
### OS

macOS 26.5.2 (Build 25F84)

### LinearMouse

v0.11.4

### Describe the bug

First of all, thank you for creating and maintaining LinearMouse. I really like this project and it has been extremely useful for me. In particular, it allows me to continue using my mouse even though its physical scroll functionality is no longer working properly. So I really appreciate the work that has gone into this project.

The Auto Scroll feature does not seem to work correctly when the initial middle-click/drag is performed over a clickable item, such as a link or a thumbnail.

For example, when using Auto Scroll in a browser, if I middle-click and start dragging directly over a link or a clickable thumbnail, Auto Scroll does not activate as expected.

A similar issue happens in Finder when trying to start Auto Scroll over a clickable item/file.

Previously, there was an option to disable the middle-click function in browsers, which I used as a workaround while browsing. However, I noticed that this option is no longer available in the current version.

### To reproduce

**Finder**

1. Enable the Auto Scroll feature in LinearMouse.
2. Open Finder and navigate to a folder containing multiple documents/files.
3. Move the mouse cursor over a clickable file/item.
4. Press the middle mouse button and immediately drag.
5. Try to use Auto Scroll.

Result: Auto Scroll does not work when the action starts directly over the clickable item.

**Browser**

1. Enable the Auto Scroll feature in LinearMouse.
2. Open a browser and navigate to a website such as YouTube.
3. Move the mouse cursor over a clickable element, for example a video thumbnail/link.
4. Press the middle mouse button and immediately drag.
5. Try to use Auto Scroll.

Result: Auto Scroll does not activate/work when the middle-click drag starts over the clickable element.

If I start the middle-click drag on an area that is not clickable, Auto Scroll works as expected.

### Expected behavior

Auto Scroll should activate regardless of whether the cursor is initially positioned over a clickable element or a non-clickable area.

Ideally, LinearMouse should intercept the middle-click + drag gesture before the underlying application handles the middle-click action, allowing Auto Scroll to work consistently.

It would also be helpful if the previous “disable middle click function in browser” option could be restored, if possible. That option was a useful workaround for this kind of conflict.

### Anything else?

The issue seems to be related to the underlying application receiving/handling the middle-click event when the cursor is over a clickable element.

The Auto Scroll feature itself works correctly when starting the gesture over non-clickable areas, so I suspect this may be related to how LinearMouse handles or intercepts middle-click events.

Thank you again for this great project. LinearMouse has been particularly useful for extending the life of my mouse despite its broken scroll functionality. I’d be happy to provide additional information, logs, or screen recordings if that would help with debugging.

gilang-gunawan:
## Summary

Auto Scroll used to feel broken in two common situations:

- Starting it over an item in Finder (or any native list/table view) had a noticeable
  lag — sometimes close to a second — before scrolling actually kicked in. It's now
  effectively instant.
- Starting it directly on a link, a button, or a file icon never worked at all — the
  click always went to the app instead. This adds an off-by-default setting to let
  Auto Scroll win in that case when you want it to.

## Why this matters

- The delay wasn't a Finder-only quirk — it sat in the code path every single Auto
  Scroll activation goes through, in any native app (Finder, Xcode, Mail, Notes, etc).
  It just wasn't obvious *why* it lagged, so it was easy to write off as "Auto Scroll is
  a bit slow" instead of catching it as a fixable bug.
- For anyone who relies on Auto Scroll as a substitute for physical scrolling (see
  #1339), a delay or an outright block on activation isn't a minor annoyance — it's the
  difference between the feature working or not.
- The click-through gap has been an open request since the original toggle for it was
  removed in #1273 for being unreliable (#1237, #1271). This closes that gap without
  reintroducing the old bug: the new setting is off by default, and unlike the old one,
  its behavior doesn't depend on click/drag history.

## Technical details

- `AutoScrollAccessibilityActivationClassifier`'s nearby-point resampling
  (`refineActivationProbe`) ran 9 extra synchronous AX hit-tests on every non-pressable
  hit, unconditionally — i.e. on every successful activation, since Auto Scroll only
  starts on non-pressable elements. That resampling exists solely to compensate for
  browser accessibility-tree imprecision near links/buttons, so it now only runs when
  the hit is actually inside web content (`AXWebArea`/`AXScrollArea` ancestor). Measured
  Finder activation latency: ~600-890ms before, ~7-120ms after.
- Added an opt-in `activateOverPressableElements` setting (default `false`) so Auto
  Scroll can start even when the initial click lands on a link, button, or clickable
  file, at the cost of the native middle-click action on that element.

## Related history

This touches the same code path as #1273, so some context on why this isn't a blind
revert of that change:

- #1103 introduced a `preserveNativeMiddleClick` toggle.
- #1237 reported that toggle was unreliable — behavior depended on whether the item had
  previously been dragged, so native middle-click sometimes silently failed.
- #1271 was a related regression (middle-click no longer closing Chrome tabs).
- #1273 fixed both by removing the toggle entirely and hard-coding "always preserve
  native click on pressable elements," explicitly to eliminate that inconsistency.
- #1317 continued hardening the same default (title-bar/tab-strip regions).

`activateOverPressableElements` is not a reintroduction of the old
`preserveNativeMiddleClick` toggle: it's additive (default `false`, so out of the box
nothing changes from #1273's behavior) rather than replacing the always-on classifier
logic, and its correctness doesn't depend on click/drag history the way the old
toggle's bug did — the classifier is a pure function of the current hit-test, not
accumulated state.

## Testing

- Full xcodebuild unit test suite (537 tests)
- SwiftFormat and SwiftLint pass
- Manually verified via `log stream` against a rebuilt/reinstalled app, not just unit
  tests:
  - Finder AX classify latency: ~600-890ms before this change, ~7-120ms after, across
    repeated activations on real Finder list rows.
  - With `activateOverPressableElements: true`: a real `AXLink` hit
    (`AXStaticText -> AXLink[press]`) still activated Auto Scroll.
  - With `activateOverPressableElements: false` (default): four separate real `AXLink`
    hits did *not* activate Auto Scroll, confirming no regression of #1273's fix.

Closes #1339

### DIFF
--- a/Documentation/Configuration.d.ts
+++ b/Documentation/Configuration.d.ts
@@ -519,6 +519,13 @@ declare namespace Scheme {
        * @description Choose the mouse button and modifier keys used to activate auto scroll.
        */
       trigger?: AutoScroll.Trigger;
+
+      /**
+       * @title Activate over clickable elements
+       * @description Activate auto scroll even if the initial click lands on a clickable element such as a link or button. When disabled (default), the click is passed through so links and buttons can still be activated normally.
+       * @default false
+       */
+      activateOverPressableElements?: boolean;
     };
 
     namespace AutoScroll {
--- a/Documentation/Configuration.json
+++ b/Documentation/Configuration.json
@@ -159,6 +159,12 @@
     "Scheme.Buttons.AutoScroll": {
       "additionalProperties": false,
       "properties": {
+        "activateOverPressableElements": {
+          "default": false,
+          "description": "Activate auto scroll even if the initial click lands on a clickable element such as a link or button. When disabled (default), the click is passed through so links and buttons can still be activated normally.",
+          "title": "Activate over clickable elements",
+          "type": "boolean"
+        },
         "enabled": {
           "default": false,
           "description": "Indicates if auto scroll is enabled.",
--- a/LinearMouse/Accessibility/AutoScrollAccessibilityActivationClassifier.swift
+++ b/LinearMouse/Accessibility/AutoScrollAccessibilityActivationClassifier.swift
@@ -44,6 +44,11 @@ struct AutoScrollAccessibilityActivationClassifier {
         "AXDisclosureTriangle",
         "AXSwitch"
     ]
+    private static let listContainerRoles: Set<String> = [
+        "AXTable",
+        "AXOutline",
+        "AXList"
+    ]
 
     private let elementQuery: AccessibilityElementQuerying
     private let bypassRuleMatcher: AccessibilityBypassRuleMatcher
@@ -76,6 +81,14 @@ struct AutoScrollAccessibilityActivationClassifier {
         }
 
         return actions.contains(kAXPressAction as String)
+    }
+
+    static func isListContainerRole(_ role: String?) -> Bool {
+        guard let role else {
+            return false
+        }
+
+        return listContainerRoles.contains(role)
     }
 
     private func refineActivationProbe(from initialProbe: AutoScrollActivationProbe) -> AutoScrollActivationProbe {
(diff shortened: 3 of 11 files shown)


lujjjh:
Thanks! There used to be a `preserveNativeMiddleClick` option, but I later removed it, and I’d prefer not to bring it back for simplicity.

Pressing and dragging over a clickable element failing to trigger Auto Scroll is a regression that needs to be fixed. The performance issue also needs to be addressed, though I’m concerned that the current optimization may stop the AX ancestor traversal too early and change the classification behavior. I’ll test and take a closer look when I have time.

PR #1341, opened 2026-08-08
Add maximize and minimize button mapping actions

DrMatrix85:
Adds two new Button Mapping actions:

Maximize Window
Minimize Window

Maximize Window resizes the focused window to the usable area of its display while preserving the menu bar and Dock, rather than entering macOS full-screen mode.

Minimize Window minimizes the focused window to the Dock using the macOS Accessibility API.

These actions work with the existing Button Mapping system, including the swipe triggers introduced in #1326.

Tested successfully with:

Logitech MX Master 4
Samsung Bluetooth Mouse Slim

Verified:

Maximize correctly fits the focused window within the visible screen area
Minimize correctly sends the focused window to the Dock
Full test suite passes
SwiftFormat lint passes

### DIFF
--- a/Documentation/Configuration.d.ts
+++ b/Documentation/Configuration.d.ts
@@ -768,8 +768,8 @@ declare namespace Scheme {
         | ShowDesktop
         | LookUpAndDataDetectors
         | SmartZoom
-	| WindowMaximize
-	| WindowMinimize
+        | WindowMaximize
+        | WindowMinimize
         | DisplayBrightnessUp
         | DisplayBrightnessDown
         | MediaVolumeUp
--- a/Documentation/Configuration.d.ts
+++ b/Documentation/Configuration.d.ts
@@ -768,6 +768,8 @@ declare namespace Scheme {
         | ShowDesktop
         | LookUpAndDataDetectors
         | SmartZoom
+	| WindowMaximize
+	| WindowMinimize
         | DisplayBrightnessUp
         | DisplayBrightnessDown
         | MediaVolumeUp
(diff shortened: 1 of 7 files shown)


lujjjh:
Thanks for the contribution! Please update `Documentation/Configuration.d.ts` with the new actions, then regenerate `Documentation/Configuration.json` by running:

```bash
npm run generate:json-schema
```

It would also be good to add the new UI strings to the localization catalog.

I’ll test it later, but the current implementation may have issues with multiple displays because AX and `NSScreen` use different coordinate systems. This could select the wrong screen or place the window incorrectly when displays have different heights or are arranged vertically.

The build also produces a new warning for the forced `AXUIElement` cast. The unchecked `AXValue` casts could similarly crash if an application returns an unexpected AX value, so these should use safe, type-checked conversions.

DrMatrix85:
Thanks for the review! I’ve addressed the requested changes:

Updated Documentation/Configuration.d.ts
Regenerated Documentation/Configuration.json
Added the new UI strings to the localization catalog
Reworked the AX/AppKit coordinate conversion for multi-display layouts
Replaced the unchecked AX casts with runtime type checks

I also rebuilt and reran the full test suite successfully, and verified Maximize/Minimize still work as expected on the hardware I have available.

lujjjh:
I’ve pushed a few follow-up changes to the branch: formatted `Configuration.d.ts`, replaced the manual string extraction with proper localization and translations, and improved window maximization reliability, particularly for Firefox, based on the approach used by Rectangle and Hammerspoon.

DrMatrix85:
Thanks for the review, follow-up improvements, and merge! I really appreciate the extra polish on the maximization behavior and localization. Glad I could contribute!

PR #1343, opened 2026-08-08
fix: accept integer-pixel scroll distances containing 0 in the JSON schema

YuriNachos:
## Summary

The JSON-schema pattern for `Scheme.Scrolling.Distance.Pixel` rejected valid integer-pixel scroll distances that contain the digit `0` (e.g. `10px`, `20px`, `100px`, `1000px`), even though the Swift runtime decoder accepts them. This made the schema stricter than the actual configuration parser, so valid configs failed schema validation / editor checks.

## Root Cause

`Documentation/Configuration.d.ts` declared the Pixel pattern as `^\d[1-9]*(\.\d+)?px`. The character class `[1-9]` excludes `0`, so any multi-digit integer containing a `0` fails to match: for `10px`, `\d` consumes `1` and `[1-9]*` cannot consume the `0`, leaving `0px` unmatched against the trailing `px`. The Swift decoder (`Distance.swift`, pattern `^([\d.]+)(px|)$`) accepts these values, so the schema was a stricter-than-runtime typo (intended `[0-9]`).

## Changes

- `Documentation/Configuration.d.ts`: Pixel `@pattern` integer body `[1-9]` → `[0-9]` (`^\d[0-9]*(\.\d+)?px`).
- `Documentation/Configuration.json`: regenerated from the type via `npm run generate:json-schema` (pre-commit hook).

Integer-pixel values (`10px`, `20px`, `100px`, `1000px`) now validate; previously-matching values (`36px`, `256px`, `3.5px`) are unchanged, and non-pixel values are still rejected.

## Verification

```
make configure clean lint test XCODEBUILD_ARGS="CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= DEVELOPMENT_TEAM="
```

- `swiftformat --lint .` clean; `swiftlint .` clean.
- `** TEST SUCCEEDED **` — 523 tests, 0 failures.
- Regex check against the generated schema pattern: `10px`/`20px`/`100px`/`1000px` now match (failed before); `36px`/`256px`/`3.5px` still match; `auto`/`abc`/`10`/`-5px`/`1.2.3px`/`px` rejected.

### DIFF
--- a/Documentation/Configuration.d.ts
+++ b/Documentation/Configuration.d.ts
@@ -240,7 +240,7 @@ declare namespace Scheme {
 
       /**
        * @description The scrolling distance in pixels.
-       * @pattern ^\d[1-9]*(\.\d+)?px
+       * @pattern ^(?:0|[1-9]\d*)(?:\.\d+)?px$
        */
       type Pixel = string;
     }
--- a/Documentation/Configuration.json
+++ b/Documentation/Configuration.json
@@ -1571,7 +1571,7 @@
     },
     "Scheme.Scrolling.Distance.Pixel": {
       "description": "The scrolling distance in pixels.",
-      "pattern": "^\\d[1-9]*(\\.\\d+)?px",
+      "pattern": "^(?:0|[1-9]\\d*)(?:\\.\\d+)?px$",
       "type": "string"
     },
     "Scheme.Scrolling.Modifiers": {


Documentation/Configuration.d.ts
```diff
@@ -240,7 +240,7 @@ declare namespace Scheme {
 
       /**
        * @description The scrolling distance in pixels.
-       * @pattern ^\d[1-9]*(\.\d+)?px
+       * @pattern ^\d[0-9]*(\.\d+)?px
```
lujjjh: The original pattern looks like it intended to reject leading zeros, but the digit classes were reversed: it accepts `01px` while rejecting `10px`. This keeps `0px` valid without allowing `01px`, and anchors the full value.

```suggestion
       * @pattern ^(?:0|[1-9]\d*)(?:\.\d+)?px$
```

YuriNachos:
Applied your suggestion — you are right on both counts, and it is a better fix than mine. Pushed as `d03f6f4` (amended, still one commit), with `Configuration.json` regenerated from the type.

I checked all three patterns against a corpus before taking it:

| value | old `^\d[1-9]*(\.\d+)?px` | mine `^\d[0-9]*(\.\d+)?px` | yours `^(?:0\|[1-9]\d*)(?:\.\d+)?px$` | Swift decoder |
| --- | --- | --- | --- | --- |
| `0px` | ✅ | ✅ | ✅ | ✅ |
| `10px` / `100px` / `1000px` | ❌ | ✅ | ✅ | ✅ |
| `36px` / `256px` / `3.5px` | ✅ | ✅ | ✅ | ✅ |
| `01px` / `007px` | ✅ | ✅ | ❌ | ✅ |
| `10pxfoo` | ✅ | ✅ | ❌ | ❌ |
| `10` / `auto` / `-5px` | ❌ | ❌ | ❌ | mixed |

Two things my version missed:

- **Leading zeros.** `01px` and `007px` passed both the old pattern and mine. Yours rejects them, which is what the reversed digit classes were evidently reaching for.
- **The missing end anchor.** Without `$`, `10pxfoo` matched. The Swift decoder (`^([\d.]+)(px|)$` in `Distance.swift`) rejects it, so anchoring actually moves the schema *closer* to runtime rather than further away.

One deliberate difference worth flagging: on `01px` the schema is now **stricter** than the decoder, which still accepts it. I read that as intended — the schema documents the canonical form — but say the word if you would rather it mirror the decoder exactly.

PR #1344, opened 2026-08-08
Scale continuous scroll deltas in pixel space in the speed-adjustment transformer

YuriNachos:
## Summary
Continuous (pixel-mode) scroll events processed by the Scrolling speed-adjustment transformer are now scaled in pixel space, fixing a ~10× under-scale for trackpad and high-resolution-wheel continuous scrolls. Discrete (line-mode) behavior is unchanged.

## Root cause
`ScrollingAccelerationSpeedAdjustmentTransformer` wrote `deltaYFixedPt = targetPt / 10` unconditionally. For discrete events `deltaYFixedPt` is in line units (10 per line), so `/ 10` is correct; but for continuous events `deltaYFixedPt` is in pixels — every other producer in the repo (`LinearScrollingVerticalTransformer`, `SmoothedScrollingTransformer`'s reader) treats it as pixels for continuous events. Dividing a pixel value by 10 left continuous scrolls ~10× too small after the speed adjustment.

## Changes
- `LinearMouse/EventTransformer/ScrollingAccelerationSpeedAdjustmentTransformer.swift` — write `deltaYFixedPt` mode-aware: `scrollWheelEventView.continuous ? targetPt : targetPt / 10` for both axes. The discrete branch is byte-for-byte the old `targetPt / 10`.
- `LinearMouseUnitTests/EventTransformer/ScrollingAccelerationSpeedAdjustmentTransformerTests.swift` (new) — continuous vertical/horizontal events keep `deltaYFixedPt` in pixel space (RED before: `0.8`, GREEN after: `8`); a discrete-event regression guard asserts the discrete path is unchanged (`1.3` before and after); an acceleration-path test confirms the matrix branch was already mode-correct.

## Test plan
- [x] `swiftformat --lint .` + `swiftlint .` — clean.
- [x] `xcodebuild test -project LinearMouse.xcodeproj -scheme LinearMouse -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO …` — 536 tests, 0 failures (the 4 new tests pass).
- [x] Convention cross-checked against the canonical reader (`SmoothedScrollingTransformer.deltaYInPixels`) and producer (`LinearScrollingVerticalTransformer`) of `deltaYFixedPt`.

Co-authored-by: Claude <noreply@anthropic.com>

### DIFF
--- a/LinearMouse/EventTransformer/ScrollingAccelerationSpeedAdjustmentTransformer.swift
+++ b/LinearMouse/EventTransformer/ScrollingAccelerationSpeedAdjustmentTransformer.swift
@@ -51,7 +51,9 @@ class ScrollingAccelerationSpeedAdjustmentTransformer: EventTransformer {
             let targetPt = scrollWheelEventView.deltaYPt + Double(deltaYSignum) * speed
             scrollWheelEventView.deltaY = deltaYSignum * max(1, Int64(abs(targetPt) / 10))
             scrollWheelEventView.deltaYPt = targetPt
-            scrollWheelEventView.deltaYFixedPt = targetPt / 10
+            // deltaYFixedPt carries pixel deltas for continuous events but line deltas
+            // for discrete ones, so keep it in the event's native delta unit.
+            scrollWheelEventView.deltaYFixedPt = scrollWheelEventView.continuous ? targetPt : targetPt / 10
             // TODO: Test if ioHidScrollY needs to be modified.
             os_log("deltaY: speed=%{public}f", log: Self.log, type: .info, speed)
         }
@@ -61,7 +63,7 @@ class ScrollingAccelerationSpeedAdjustmentTransformer: EventTransformer {
             let targetPt = scrollWheelEventView.deltaXPt + Double(deltaXSignum) * speed
             scrollWheelEventView.deltaX = deltaXSignum * max(1, Int64(abs(targetPt) / 10))
             scrollWheelEventView.deltaXPt = targetPt
-            scrollWheelEventView.deltaXFixedPt = targetPt / 10
+            scrollWheelEventView.deltaXFixedPt = scrollWheelEventView.continuous ? targetPt : targetPt / 10
             os_log("deltaX: speed=%{public}f", log: Self.log, type: .info, speed)
         }
 
--- a/LinearMouseUnitTests/EventTransformer/ScrollingAccelerationSpeedAdjustmentTransformerTests.swift
+++ b/LinearMouseUnitTests/EventTransformer/ScrollingAccelerationSpeedAdjustmentTransformerTests.swift
@@ -1,0 +1,166 @@
+// MIT License
+// Copyright (c) 2021-2026 LinearMouse
+
+@testable import LinearMouse
+import XCTest
+
+final class ScrollingAccelerationSpeedAdjustmentTransformerTests: XCTestCase {
+    /// The fixed-point delta fields are stored as 16.16 fixed point; allow for one
+    /// quantization step when comparing against the expected value.
+    private let fixedPointTolerance = 1.0 / 32_768.0
+
+    func testSpeedAdjustmentKeepsContinuousFixedPointInPixelSpace() throws {
+        // A continuous (pixel-mode) scroll event carries its delta in pixel units: both
+        // deltaYPt and deltaYFixedPt hold the same pixel value.
+        let transformer = ScrollingAccelerationSpeedAdjustmentTransformer(
+            acceleration: Scheme.Scrolling.Bidirectional<Decimal>(vertical: nil, horizontal: nil),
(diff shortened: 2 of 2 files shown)


lujjjh:
Could you describe the concrete user-visible issue that led to this change, including the device, scrolling mode/settings, affected application, and reproduction steps? Native pixel-unit CGEvents use fixedPtDelta = pointDelta / 10, so it is unclear which real producer or consumer expects these fields to be equal. Was this behavior observed on actual events, or inferred solely from the existing code?

AmorphousSheep:
I think this relates to an issue I was going to raise where with a high res wheel and smooth scrolling enabled, the smallest tick of the wheel moves too great a distance for true smoothness at slow scroll speeds.

YuriNachos:
Straight answer to your last question first: **inferred from the existing code, not observed on real events.** I don't have a capture of native CGEvents to show you, and I should have said that in the PR body rather than leaving it to be asked.

What I can point at is a real consumer *and* a real producer inside this repo — which is what I was going on.

**Consumer** — `SmoothedScrollEventDelivery.deltaYInPixels`, `LinearMouse/EventTransformer/SmoothedScrollingTransformer.swift:518-536`:

```swift
if !view.continuous {
    ...
    if view.deltaYFixedPt != 0 {
        return view.deltaYFixedPt * Self.inputLineStepInPoints * 10   // line units → pixels
    }
}
if view.deltaYFixedPt != 0 {
    return view.deltaYFixedPt                                        // continuous: already pixels
}
```

For a **discrete** event it converts `deltaYFixedPt` to pixels via `* 36 * 10`, i.e. it reads the field as tenths of a line. For a **continuous** event it falls through and returns `deltaYFixedPt` **unscaled, as pixels**. Those two branches only make sense together if continuous `fixedPtDelta` is in pixel space.

**Producer** — `setVertical`, same file, `:546-551`:

```swift
func setVertical(_ value: Double, on view: ScrollWheelEventView) {
    view.deltaY = integerDelta(for: value)
    view.deltaYPt = pointDeltaAccumulator.verticalPointDelta(for: value)
    view.deltaYFixedPt = value       // `value` is in pixels
    view.ioHidScrollY = value
}
```

When LinearMouse emits a smoothed continuous event it writes the pixel value straight into `deltaYFixedPt`, not `value / 10`.

So the concrete thing this PR is reacting to is that **the repo currently holds both positions at once**: `SmoothedScrollEventDelivery` treats continuous `fixedPtDelta` as pixels in both directions, while `ScrollingAccelerationSpeedAdjustmentTransformer` writes `targetPt / 10` into it for every event, continuous or not. One of those is wrong regardless of which way it resolves, and they should agree.

**If you're right that native pixel-unit events use `fixedPtDelta = pointDelta / 10`** — and you know this hardware far better than I do — then the bug is on the other side: `deltaYInPixels`'s continuous branch and `setVertical` are the ones misreading the unit, and this PR is aimed at the wrong file. In that case I'd rather close it and open an issue against the smoothed-scrolling path than have you spend more review on it. Say the word and I'll do that.

**If you'd like it settled empirically rather than by reading**, I'm happy to do the measurement instead of guessing: a small `CGEventTap` logger dumping `deltaAxis1` / `pointDeltaAxis1` / `fixedPtDeltaAxis1` / `isContinuous` for a trackpad, a Magic Mouse, a plain wheel and a Logitech high-resolution wheel, and post the observed ratios. That answers "what do real producers actually emit" directly, and it would be worth having on record whichever way it comes out. It just needs the hardware in front of me, so tell me which devices matter most to you.

I'd rather get this right than get it merged. On @AmorphousSheep's note — a high-res wheel plus smooth scrolling moving too far on the smallest tick sounds plausibly related, but it runs through `SmoothedScrollingTransformer` and the high-resolution normalizer rather than this transformer, so I don't want to claim it as a repro for this change without measuring it first.

YuriNachos:
Following up on my note above — the two ways this could resolve (this transformer as written, or the smoothed-scrolling side being the real unit mismatch) are both still open from my side. If the empirical route would settle it, tell me which devices matter most and I'll post the observed fixedPtDelta / pointDelta / isContinuous ratios from a CGEventTap logger; and if you'd rather I close this and re-file against the SmoothedScrollingTransformer path, I'm glad to do that instead. Whichever way you lean, I'll take it from there.

lujjjh:
Thanks for following up. I captured native continuous trackpad events and found that `fixedPtDelta` is not equal to `pointDelta`; their relationship also varies between events.

Since this PR is based on an inferred invariant rather than a reproducible user-facing issue, and its tests assume that invariant in their inputs, I don’t think we should change these fields by roughly 10× without stronger empirical evidence.

The handling in `SmoothedScrollingTransformer` may still deserve investigation, but it should start from a concrete issue or additional measurements. I’m going to close this PR for now. Thanks for identifying the inconsistency and discussing it carefully.

PR #1354, opened 2026-08-10
Accept a zero gesture dead-zone instead of disabling all gestures

YuriNachos:
## Summary
`GestureButtonTransformer.detectGesture` guarded the axial-tolerance check with strict `< deadZone`, so a valid `deadZone` of 0 (`@minimum 0`, default 40) made `absDelta < 0` always false and disabled every side-button gesture. A 0 dead zone should mean "only perfectly axial gestures are accepted", not "gestures off".

## Root Cause
`LinearMouse/EventTransformer/GestureButtonTransformer.swift`, `detectGesture`: strict `<` instead of inclusive `<=` on the dead-zone comparison (the file already uses inclusive `magnitude >= threshold` comparisons elsewhere).

## Changes
- `< deadZone` -> `<= deadZone` on the Y (and X) dead-zone comparisons in `detectGesture`.
- `GestureButtonTransformerTests`: a perfectly axial drag at `deadZone == 0` is now detected (red before, green after); an off-axis drag is still rejected at `deadZone == 0`.

## Verification
- `make lint` — no new violations vs main.
- `GestureButtonTransformerTests` passes (targeted run, headless, code signing disabled). CI runs the full suite.

## Notes
1-px boundary shift for non-zero dead zones (negligible, strictly more correct: a tolerance of N px accepts drift 0..N inclusive).

### DIFF
--- a/LinearMouse/EventTransformer/GestureButtonTransformer.swift
+++ b/LinearMouse/EventTransformer/GestureButtonTransformer.swift
@@ -249,7 +249,7 @@ extension GestureButtonTransformer: EventTransformer {
         // Determine dominant axis
         if absDeltaX > absDeltaY {
             // Horizontal gesture
-            guard absDeltaY < deadZone else {
+            guard absDeltaY <= deadZone else {
 //                os_log(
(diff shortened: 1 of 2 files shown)


lujjjh:
Thanks!

PR #1360, opened 2026-08-14
Debounce phantom mouse-up events in the legacy click debouncer

### LINKED ISSUE #1316: debounce feature does not work properly (including proposed solution)
### OS

macOS 26.5.2 (Build 25F84)

### LinearMouse

vsha-3d9ba49

### Describe the bug

My Evoluent vertical mouse 4 bounces the left button creating annyoing unwanted double clicks. Using the debounce feature of LinearMouse gives no relief.
After studying the coding of class ClickDebouncingTransformer I suspect the problem is: Only Down events are checked, but not Up events. 
On my Mac I locally changed and tried the code below, and now it seems to work properly, no more unwanted double clicks recently.
Please consider taking over these or similar changes. I don't dare committing myself, as I have no experience with this procedure.
Also the log shows that more up events are skipped than down events.

class ClickDebouncingTransformer: EventTransformer {
    private static let log = OSLog(subsystem: Bundle.main.bundleIdentifier!, category: "ClickDebouncing")

    private let button: CGMouseButton
    private let timeout: TimeInterval
    private let resetTimerOnMouseUp: Bool

    private var debouncedDownClicks: UInt64 = 0
    private var debouncedUpClicks: UInt64 = 0

    init(for button: CGMouseButton, timeout: TimeInterval, resetTimerOnMouseUp: Bool) {
        self.button = button
        self.timeout = timeout
        self.resetTimerOnMouseUp = resetTimerOnMouseUp
    }

    private var mouseDownEventType: CGEventType {
        button.fixedCGEventType(of: .leftMouseDown)
    }

    private var mouseUpEventType: CGEventType {
        button.fixedCGEventType(of: .leftMouseUp)
    }

    private var lastClickedDownAtInNanoseconds: UInt64 = 0
    private var lastClickedUpAtInNanoseconds: UInt64 = 0

    func transform(_ event: CGEvent, in _: EventTransformerContext) -> CGEvent?{
        guard [mouseDownEventType, mouseUpEventType].contains(event.type) else {
            return event
        }
        let mouseEventView = MouseEventView(event)
        guard mouseEventView.mouseButton == button else {
            return event
        }
        
        switch event.type {
        case mouseDownEventType:
            let intervalSinceLastClickDown = intervalSinceLastClickDown
            touchLastClickedDownAt()
            if intervalSinceLastClickDown <= timeout {
                debouncedDownClicks += 1
                os_log(
                    "Mouse down ignored because interval since last click %{public}f <= %{public}f, #:%{public}%llu",
                    log: Self.log,
                    type: .info,
                    intervalSinceLastClickDown,
                    timeout,
                    debouncedDownClicks
                )
                return nil
            }
            return event
        case mouseUpEventType:
            let intervalSinceLastClickUp = intervalSinceLastClickUp
            touchLastClickedUpAt()
            if intervalSinceLastClickUp <= timeout {
                debouncedUpClicks += 1
                os_log(
                    "Mouse up ignored because interval since last click %{public}f <= %{public}f, #:%{public}%llu",
                    log: Self.log,
                    type: .info,
                    intervalSinceLastClickUp,
                    timeout,
                    debouncedUpClicks
                )
                return nil
            }
            if resetTimerOnMouseUp {
                touchLastClickedDownAt()
            }
            return event
        default:
            break
        }

        return event
    }

    private func touchLastClickedDownAt() {
        lastClickedDownAtInNanoseconds = DispatchTime.now().uptimeNanoseconds
    }

    private var intervalSinceLastClickDown: TimeInterval {
        let nanosecondsPerSecond = 1e9
        return Double(DispatchTime.now().uptimeNanoseconds - lastClickedDownAtInNanoseconds) / nanosecondsPerSecond
    }
    private func touchLastClickedUpAt() {
        lastClickedUpAtInNanoseconds = DispatchTime.now().uptimeNanoseconds
    }

    private var intervalSinceLastClickUp: TimeInterval {
        let nanosecondsPerSecond = 1e9
        return Double(DispatchTime.now().uptimeNanoseconds - lastClickedUpAtInNanoseconds) / nanosecondsPerSecond
    }
}

### To reproduce

Reproducing will be difficult for anyone not having a defective mouse...and maybe other mice bounce differently. For testing I used [https://debouncetester.com/](url)

### Expected behavior

No more annoying double clicks

### Anything else?

_No response_

YuriNachos:
## Summary

The legacy click debouncer (`ClickDebouncingTransformer`) only debounced *presses*. A bouncing mouse switch also produces phantom *releases* after the real one (Real Down → Real Up → Bounce Down → Bounce Up); the bounce press was suppressed but the bounce release leaked through to the app, so unwanted double-clicks still registered. This gives mouse-up events the same sliding debounce window mouse-down events already had.

Closes #1316

## Root Cause

In `ClickDebouncingTransformer.transform`, the `mouseUpEventType` case never consulted a release timestamp: it only optionally reset the press window (`resetTimerOnMouseUp`) and always forwarded the event. For the reporter's measured pattern the bounce press was suppressed by the press-side window, but the bounce release reached the app — and in the Down/Up/Down/Up bounce variant both extra events leaked. The reporter's log shows more up events skipped than down events reaching the app, matching a release-side leak.

## Changes

- Track `lastReleasedAtInNanoseconds` and suppress a release when `intervalSinceLastRelease <= timeout`, mirroring the press-side logic line-for-line (including the `os_log` wording). The window slides: the timestamp updates on every release event before the window check, same ordering as the press side.
- A release is only ever suppressed when it does not close a forwarded press (`unreleasedPressCount == 0`): a forwarded press must always get its forwarded release, or apps would see a mouse down without its mouse up (stuck button / drag glue). This mirrors the pairing invariant the project's own `LibinputClickDebouncingTransformer` upholds.
- `resetTimerOnMouseUp` now only resets the press window for *forwarded* releases: a suppressed release is contact bounce, not a real release, so it does not re-arm the press debounce window. This confines the observable change to "phantom releases no longer reach the app".

## Verification

`make configure clean lint test XCODEBUILD_ARGS="CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= DEVELOPMENT_TEAM="` (steps run individually in a worktree — `make configure`'s `.git/hooks` target cannot run where `.git` is a file; same commands, same arguments):

- swiftformat: 0/253 files require formatting; swiftlint: 0 serious violations (none in the changed files)
- **563 tests, 0 failures**, including 10 new/updated tests in `ClickDebouncingTransformerTests`, all proven red on the pre-fix code:
  - the reporter's exact pattern (bounce up after suppressed bounce down) — suppressed
  - intentional rapid clicking (60 ms cadence vs 50 ms window) — still forwarded
  - the release that closes a forwarded press inside the window — always forwarded (stuck-button regression)
  - real release after bounce chatter at timeout=100 ms (#1316's worn-switch audience) — forwarded
  - suppressed release does not re-arm the press window; window anchors at last release; `<=` boundary; trailing-edge extension

## Known limitation

The down/up pair can in principle be split across transformer routes: `EventTransformerManager` re-resolves the route per event and only pins a button stream when an `EventTransformerInteractionTracking` transformer claimed it — the legacy click debouncer is not one. So with legacy debounce enabled, a press routed via window A and its release via window B (a drag across windows/screens, with a large timeout and a recently forwarded release on route B) could have that release suppressed by route B's instance, leaving the app with a mouse-down without its mouse-up. This needs an unusual combination of circumstances, and pre-fix code masked it only by never suppressing releases at all. The clean fix is upstream route-pinning of down/up pairs (adopting `EventTransformerInteractionTracking` for the debouncer); this PR keeps to the debouncer itself.

### DIFF
--- a/LinearMouse/EventTransformer/ClickDebouncingTransformer.swift
+++ b/LinearMouse/EventTransformer/ClickDebouncingTransformer.swift
@@ -33,6 +33,12 @@ class ClickDebouncingTransformer: EventTransformer {
     }
 
     private var lastClickedAtInNanoseconds: UInt64 = 0
+    private var lastReleasedAtInNanoseconds: UInt64 = 0
+
+    /// Forwarded presses that have not been closed by a forwarded release yet.
+    /// A counter rather than a flag: event streams from other applications are
+    /// not guaranteed to alternate down/up.
+    private var unreleasedPressCount = 0
 
     func transform(_ event: CGEvent, in _: EventTransformerContext) -> CGEvent? {
         guard [mouseDownEventType, mouseUpEventType].contains(event.type) else {
@@ -57,10 +63,34 @@ class ClickDebouncingTransformer: EventTransformer {
                 )
                 return nil
             }
+            unreleasedPressCount += 1
             return event
         case mouseUpEventType:
+            // Never swallow the release that closes a forwarded press: the app
+            // would see a mouse down without its mouse up and treat the button
+            // as stuck. Only releases with no forwarded press to close are
+            // bounce artifacts and may be dropped inside the window.
+            let intervalSinceLastRelease = intervalSinceLastRelease
+            touchLastReleasedAt()
+            // A suppressed release is still a physical contact break: the next
+            // bounce-down must land inside the press window, so the press timer
+            // re-arms on every release, forwarded or not — the same sliding
+            // behavior the press side applies to suppressed presses.
             if resetTimerOnMouseUp {
                 touchLastClickedAt()
+            }
+            if intervalSinceLastRelease <= timeout, unreleasedPressCount == 0 {
+                os_log(
+                    "Mouse up ignored because interval since last release %{public}f <= %{public}f",
+                    log: Self.log,
+                    type: .info,
+                    intervalSinceLastRelease,
+                    timeout
+                )
+                return nil
+            }
+            if unreleasedPressCount > 0 {
+                unreleasedPressCount -= 1
             }
             return event
         default:
@@ -74,8 +104,17 @@ class ClickDebouncingTransformer: EventTransformer {
         lastClickedAtInNanoseconds = nowInNanoseconds()
     }
 
+    private func touchLastReleasedAt() {
+        lastReleasedAtInNanoseconds = nowInNanoseconds()
+    }
+
     private var intervalSinceLastClick: TimeInterval {
         let nanosecondsPerSecond = 1e9
         return Double(nowInNanoseconds() - lastClickedAtInNanoseconds) / nanosecondsPerSecond
+    }
+
+    private var intervalSinceLastRelease: TimeInterval {
+        let nanosecondsPerSecond = 1e9
+        return Double(nowInNanoseconds() - lastReleasedAtInNanoseconds) / nanosecondsPerSecond
     }
 }
--- a/LinearMouseUnitTests/EventTransformer/ClickDebouncingTransformerTests.swift
+++ b/LinearMouseUnitTests/EventTransformer/ClickDebouncingTransformerTests.swift
@@ -93,7 +93,7 @@ private func makeMouseButtonEvent(
 }
 
 final class ClickDebouncingTransformerTests: XCTestCase {
-    func testLegacyModeSuppressesRapidPressButAlwaysForwardsRelease() throws {
+    func testLegacyModeSuppressesRapidPressButForwardsFirstRelease() throws {
         var now: UInt64 = 1_000_000_000
         let transformer = ClickDebouncingTransformer(
             for: .left,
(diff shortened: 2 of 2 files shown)


lujjjh:
Thanks for working on this. The legacy debouncer has several known limitations because its handling of mouse-down and mouse-up events is intentionally very simple. In particular, the same down/up/down/up sequence can represent different physical bounce patterns, which cannot be distinguished reliably from timing alone. Suppressing mouse-up events may also reintroduce the stuck drag and resize issues fixed in #1084.

These limitations are why I introduced the libinput-based debouncing mode. I would prefer not to add further complexity or behavioral changes to the legacy implementation, and instead focus future fixes on the newer mode.

For that reason, I’m going to close this PR. Thank you for investigating the issue and adding the tests.

PR #1361, opened 2026-08-15
Add media play/pause, next, back to gestures

ctlee:
Expanding gesture control options to include media play/pause, next, and back.

### DIFF
--- a/Documentation/Configuration.d.ts
+++ b/Documentation/Configuration.d.ts
@@ -1308,7 +1308,10 @@ declare namespace Scheme {
         | "missionControl"
         | "appExpose"
         | "showDesktop"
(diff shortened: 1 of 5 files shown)


lujjjh:
Thanks for the contribution. Button Mappings now supports swipe gestures and already provides media actions such as play/pause, next, and previous.

I plan to move the standalone Gesture Button feature into Button Mappings instead of maintaining two separate gesture and action systems. Extending the existing Gesture Button action list would duplicate functionality and create more maintenance work during that transition.

For that reason, I don’t plan to merge this change, but the requested behavior can already be configured through Button Mappings. Thanks again for taking the time to implement it.

Issue #1177, opened 2026-05-03
Mac wakes up from sleep


Bonez76:
### OS

MacOS 26.4.1

### LinearMouse

0.11.1

### Describe the bug

After I have updated to 0.11.1 yesterday, my Mac Mini M2 wakes up from sleep every time, although I don´t do anything. 

This does not happen, when my Bluetooth mouse (Logi M650 L) is turned off or LinearMouse app is closed. Once I start LinearMouse again, sleep modes get interrupted again. 

Ran "pmset -g log | grep Wake" and got this:
`DarkWake to FullWake from Deep Idle [CDNVA] : due to HID Activity Using AC (Charge:0%)`

[LinearMouse.log](https://github.com/user-attachments/files/27314325/LinearMouse.log)

### To reproduce

1. Run "LinearMouse"
2. Open "Apple" menu and select "Sleep Mode"
3. Wait until monitor shows now signal
4. immediately after "no signal", Mac wakes up

### Expected behavior

Mac not waking up

lujjjh:
Thanks! I can reproduce it. This will be fixed in the next beta release.

PR #1364, opened 2026-08-18
Re-apply pointer settings while the system settles after waking

### LINKED ISSUE #1047: LinearMouse - stops working when 0s 26.2 goes to sleep
### OS

macOS 26.2 (Build 25C56)

### LinearMouse

v0.10.2

### Describe the bug

Hi — whenever my MacBook goes to sleep, then I wake it up, I have to relaunch linear mouse. Have you any idea how to solve this, please? Thanks

### To reproduce

_No response_

### Expected behavior

_No response_

### Anything else?

_No response_

### LINKED ISSUE #1204: LinearMouse not working after wake from standby
### OS

MacOS 26.3.1

### LinearMouse

v0.11.2

### Describe the bug

When my computer wakes from standby, LinearMouse seems inactive until I open the "LinearMouse Settings" panel.

### To reproduce

1. Let the computer go into standby.
2. Wake the computer up.
3. Observe mouse is slow without LinearMouse acceleration and speed settings.
4. Click on LinearMouse icon → "LinearMouse Settings" and suddenly the mouse is running at my chosen speed (without editing any settings, just opening the panel).

### Expected behavior

LinearMouse settings still apply after wake from standby.

### Anything else?

1. I'm using a Logitech MX Master 3 and an Apple Magic Trackpad.
2. The mouse is connected with a Logitech Unifying Receiver plugged into a USB-C dock (I'd prefer not to change this because the dock is an iDock M10, also a KVM switch).

Possibly related to #1177 or an unintended side effect of the recent bugfix.

ksc98:
## The bug

After the Mac wakes from sleep, the pointer runs on macOS's own acceleration curve instead of the configured scheme. With `"pointer": { "disableAcceleration": true }` this is very noticeable — the mouse feels slow and mushy at low speeds. The app is still running and its UI works fine.

It fixes itself the moment something unrelated happens. From #1204:

> LinearMouse seems inactive until I open the "LinearMouse Settings" panel. [...] Observe mouse is slow without LinearMouse acceleration and speed settings.

and #1047:

> On wakeup, mouse has reverted to MacOS acceleration (i.e. very slow). The mouse acceleration stays slow **until I click on the desktop**. After doing that the mouse returns to being accelerated by LinearMouse. This is reproducible every time.

Fixes #1204. Very likely also fixes #1047 (same symptom and same self-heal trigger, though that one was reported against v0.10.2, before the regression below existed).

## The fix

- Add `WakeReapplyScheduler`, which re-runs a block once a second for 30 seconds and cancels the pending series when another wake arrives.

The cadence is flat rather than backed off. The reset can land anywhere in the settle window, so the interval is the worst case for how long the pointer stays wrong before it is corrected — a backoff would leave a multi-second gap late in the window, which is exactly when a slow receiver or a Bluetooth reconnect finishes. A flat 1s poll is also the shape `AccessibilityPermission.pollingUntilEnabled` already uses. `updatePointerSpeed()` only writes IOKit properties and already runs on every frontmost-app change, so 30 extra calls spread over 30 seconds is well inside normal operating load.
- `DeviceManager.reapplyPointerSpeedAfterWake()` uses it to re-run `updatePointerSpeed()`, and `DeviceManager.stop()` cancels any pending series so nothing fires while the app is stopped.
- `AppDelegate` calls it from the `didWake` handler, next to the existing `requestLogitechControlsReconfigurationAfterWake()`.

`updateHardwareDPI()` and `updateHighResolutionWheel()` are deliberately left out of the repeated pass: they send real HID++ traffic to the device, and they already have their own post-wake and reconnect handling from #1256. `updatePointerSpeed()` is idempotent and only writes IOKit properties, so re-running it a few times costs nothing.

## Why it happens

Waking is not a single moment. `NSWorkspace.didWakeNotification` fires before the HID stack has finished restoring device properties. `AppDelegate` handles it synchronously — `restartIfAllowed()` → `DeviceManager.start()` → `updatePointerSpeed()` — so LinearMouse writes `HIDPointerResolution` / `HIDUseLinearScalingMouseAcceleration` into a device that is not ready to keep them, and the values are gone moments later. Nothing re-applies them.

What makes this stick rather than self-correct is that the app has already put the device back on the macOS defaults itself. #1182 added a `willSleepNotification` handler that calls `stop()`, and `DeviceManager.stop()` calls `restorePointerSpeedToInitialValue()`. So the sequence across a sleep is:

1. `willSleep` → `stop()` → every device is written back to the macOS system values.
2. `didWake` → `start()` → the configured values are written straight back, too early to survive.
3. Nothing else runs, so the device keeps the system values from step 1.

Before #1182 there was no sleep or wake handling at all, so step 1 never happened and the configured values simply stayed on the device across most sleeps. That makes this a regression in v0.11.2 — and it matches #1204, which was filed against v0.11.2 and guessed as much ("Possibly related to #1177 or an unintended side effect of the recent bugfix").

The recovery in both reports is incidental. Every other caller of `updatePointerSpeed()` is an unrelated trigger:

| Trigger | Wired in |
| --- | --- |
| frontmost app changed | `didActivateApplicationNotification` observer |
| config file edited | `ConfigurationState.$configuration` sink |
| current display changed | `ScreenManager.$currentScreenName` sink |
| device (re)appeared | `deviceAdded` |

Clicking the desktop activates Finder (row 1). Opening the settings window activates LinearMouse (row 1). A device that happens to re-enumerate across sleep recovers via row 4, which is why the bug looks intermittent and hardware-dependent.

The property observers registered in `DeviceManager.init` do not cover this case: `IOHIDEventSystemClientRegisterPropertyChangedCallback` watches event-system-level properties — what System Settings → Mouse writes — not the per-service `HIDEventServiceProperties` that get reset here.

This is the same shape as #1256, which fixed the high-resolution wheel not surviving wake by retrying while the HID++ feature was unavailable. Pointer speed and acceleration were the remaining piece.

## Test plan

- `make lint` (swiftformat + swiftlint) passes.
- `make test` passes; `WakeReapplySchedulerTests` adds 4 cases covering the schedule, coalescing of overlapping wakes, cancellation, and the shape of the default delays.
- Manual, on macOS 26.5 with a Logitech Unifying receiver behind a Thunderbolt dock and a scheme setting `pointer.disableAcceleration: true`. Read the applied value back from the pointer event service (the receiver publishes two; the pointer one carries `HIDPointerAccelerationType = HIDMouseAcceleration`):

  ```
  ioreg -r -c IOHIDEventService -l | grep -o 'HIDUseLinearScalingMouseAcceleration"=[0-9]*'
  ```

  Before: after wake the value reads `0` and stays there until the frontmost app changes. After: it returns to `1` within the settle window without any interaction.

### DIFF
--- a/LinearMouse/AppDelegate.swift
+++ b/LinearMouse/AppDelegate.swift
@@ -127,6 +127,7 @@ extension AppDelegate {
             os_log("System did wake", log: Self.log, type: .info)
             self?.sleeping = false
             self?.restartIfAllowed()
+            self?.reapplyPointerSpeedAfterWake()
             self?.requestLogitechControlsReconfigurationAfterWake()
         }
     }
@@ -142,6 +143,17 @@ extension AppDelegate {
     func restartIfAllowed() {
         stop()
         startIfAllowed()
+    }
+
+    /// `restartIfAllowed()` applies the pointer settings as soon as the wake notification arrives,
+    /// which is before macOS has finished restoring its own HID properties. Keep re-applying them
+    /// until the system has settled.
+    func reapplyPointerSpeedAfterWake() {
+        guard sessionActive, !sleeping else {
+            return
+        }
+
+        DeviceManager.shared.reapplyPointerSpeedAfterWake()
     }
 
     func requestLogitechControlsReconfigurationAfterWake() {
--- a/LinearMouse/Device/DeviceManager.swift
+++ b/LinearMouse/Device/DeviceManager.swift
@@ -14,6 +14,7 @@ class DeviceManager: ObservableObject {
 
     private let manager = PointerDeviceManager()
     private let receiverMonitor = ReceiverMonitor()
+    private let wakeReapplyScheduler = WakeReapplyScheduler()
 
     private var pointerDeviceToDevice = [PointerDevice: Device]()
     @Published private(set) var receiverPairedDeviceIdentities = [Int: [ReceiverLogicalDeviceIdentity]]()
@@ -70,6 +71,8 @@ class DeviceManager: ObservableObject {
     private var activateApplicationObserver: Any?
 
     func stop() {
+        wakeReapplyScheduler.cancel()
+
         guard state == .running else {
             return
         }
@@ -384,6 +387,26 @@ class DeviceManager: ObservableObject {
         }
 
         device.applyConfiguredHighResolutionWheel(highResolutionWheel)
+    }
+
+    /// Re-applies the pointer settings across the window in which macOS restores its own HID
+    /// properties after a wake.
+    ///
+    /// The settings applied while handling `NSWorkspace.didWakeNotification` are overwritten by the
+    /// system shortly afterwards, leaving the pointer on the macOS defaults until something
+    /// unrelated (a frontmost-app change, a configuration edit) happens to call
+    /// `updatePointerSpeed()` again. See `WakeReapplyScheduler`.
+    func reapplyPointerSpeedAfterWake() {
+        os_log(
+            "Re-applying pointer settings for the %{public}.0fs after wake",
+            log: Self.log,
+            type: .info,
+            wakeReapplyScheduler.duration
+        )
+
+        wakeReapplyScheduler.restart { [weak self] in
+            self?.updatePointerSpeed()
+        }
     }
 
     func restorePointerSpeedToInitialValue() {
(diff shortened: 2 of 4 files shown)


ksc98:
Closing in favour of a fresh PR with the full root-cause analysis (this one predates finding that the pointer's IOHIDEventService is terminated and re-created across sleep).

PR #1369, opened 2026-08-20
Support matching bundle-less processes by executable name in the app picker

Outfire:
Hi! A short intro on how I ended up here. I use LinearMouse daily, but I have zero Swift experience so all changes made by Claude. Recently I started running a Windows applications through CrossOver (Wine) on my Mac and wanted per-app settings for it and found there was no way to configure them from the GUI at all.

Summary

Wine/CrossOver/Whisky processes have no bundle identifier, so they never show up in the app picker. "Other Executable…" does not help either: Wine copies the loader into a per-launch temporary directory (/var/folders/…/winetemp-<random>/Some.exe), so an exact path stops matching on the next launch.

Wine/CrossOver applications and other bundle-less processes have no bundle identifier, and their executable path points into a per-launch temporary directory, so neither existing app picker option can target them. The matching engine already supports if.processName; this exposes it in the GUI:

- Running bundle-less processes are listed in the app picker by executable name, alongside running apps.
- Schemes configured with if.processName appear in the Configured section and survive GUI edits.

Covered by a scheme-index test asserting processName schemes are found without cross-matching app or processPath conditions.


lujjjh:
Thanks!

PR #1370, opened 2026-08-20
Post simulated key events with HID system state to keep input state intact

Outfire:
This is the follow-up fix. Similar situation to [#1369](https://github.com/linearmouse/linearmouse/pull/1369).
In this case CrossOver + MX Master 4. Side wheel remapped to keyboard buttons F1/F2. If hold RBM and spin side wheel for triggering F1/F2, RBM still in hold state but no action until release and press it again.

Summary

Key events created without an event source carry no HID system state, a zero timestamp and no keyboard type. Some event consumers treat such events as foreign and mistrack input state around them - e.g. Wine applications drop a held-button mouse capture when a key press arrives, and Wine's modifier tracking leaves Control stuck after a space-switch symbolic hotkey (Control+Arrow) fires mid focus change, turning later game input into Control-modified keys. Create simulated events (both KeySimulator key presses and symbolic hotkey sequences) from a hidSystemState source and stamp a real timestamp and keyboard type so they are indistinguishable from hardware key presses.

Covered by a KeySimulator test recording posted events and asserting the HID system state, timestamp and keyboard type fields.

### DIFF
--- a/Modules/KeyKit/Sources/KeyKit/CGEvent+HardwareLikeKeyEvent.swift
+++ b/Modules/KeyKit/Sources/KeyKit/CGEvent+HardwareLikeKeyEvent.swift
@@ -1,0 +1,24 @@
+// MIT License
+// Copyright (c) 2021-2026 LinearMouse
+
+import CoreGraphics
+import Foundation
+
+extension CGEvent {
+    /// Creates a key event that is indistinguishable from a hardware key press.
+    ///
+    /// Events created without a source carry no HID system state, a zero timestamp and no keyboard type.
+    /// Some event consumers (e.g. Wine's mouse-capture and modifier tracking in games) treat such events
+    /// as foreign and mistrack input state around them. The keyboard type is read from the event source
+    /// rather than `LMGetKbdType()`, which is not thread-safe and would race with `KeyCodeResolver`.
+    static func makeHardwareLikeKeyEvent(virtualKey: CGKeyCode, keyDown: Bool) -> CGEvent? {
+        guard let source = CGEventSource(stateID: .hidSystemState),
+              let event = CGEvent(keyboardEventSource: source, virtualKey: virtualKey, keyDown: keyDown) else {
+            return nil
+        }
+
+        event.timestamp = CGEventTimestamp(DispatchTime.now().uptimeNanoseconds)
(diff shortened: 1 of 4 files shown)


lujjjh:
Thanks!

One concern is that Apple documents `LMGetKbdType()` as not thread-safe. Its existing use in `KeyCodeResolver` is confined to the main thread, but the new calls in `KeySimulator` and `makeHardwareLikeKeyEvent()` can run on LinearMouse’s `EventThread` while other calls run on main. The per-instance `KeySimulator` locks also do not synchronize with each other or with the symbolic-hotkey path.

One possible direction would be to introduce a small shared keyboard-type snapshot in `KeyKit`. `KeyCodeResolver.updateMapping()` already runs on the main thread, so it could read `LMGetKbdType()` there and publish the value under a lock. `KeySimulator` and `SymbolicHotKey` would then only read that snapshot when constructing events. The snapshot could be refreshed through the existing input-source update path, rather than calling the Carbon API for every event or treating the value as a process-lifetime constant.

Outfire:
Thanks, good catch. Resolved by removing the `LMGetKbdType()` calls from the event path entirely rather than adding a snapshot.

Each simulated event is now created from its own `CGEventSource(stateID: .hidSystemState)`, and the keyboard type is read from that source's `keyboardType`, a CoreGraphics accessor. I verified it returns the same value as `LMGetKbdType()` both on and off the main thread. The only remaining `LMGetKbdType()` call is the existing one in `KeyCodeResolver`, still main-thread only, so there is no cross-thread access left to synchronize.

I went this way instead of the snapshot for three reasons. There is no shared mutable state, so nothing for the `KeySimulator` instances or the symbolic-hotkey path to lock against. The value is fresh per event, whereas a snapshot refreshed on input-source notifications would not update when the keyboard hardware changes. And `postSymbolicHotKey` can fire from the event thread before any `KeySimulator` (and therefore any `KeyCodeResolver`) exists, so the snapshot would have needed an initial main-thread hop from the event thread, which risks a deadlock against `EventThread.performAndWait`.

Creating a source per event is cheap at key-press rates. One caveat found while testing: `CGEvent(keyboardEventSource:)` does not reliably inherit the source's keyboard type (in the xctest process the source reported 47 but the event got 92), so the explicit `keyboardEventKeyboardType` stamp stays.

lujjjh:
Looks good to me. Thanks for the update!

PR #1375, opened 2026-08-30
Honor XDG_CONFIG_HOME when resolving the configuration path

Mittelwelt:
## Motivation

LinearMouse hardcodes `~/.config/linearmouse/linearmouse.json` and ignores `XDG_CONFIG_HOME`. `~/.config` is *the* XDG Base Directory default, so on a machine that relocates its XDG config root LinearMouse still recreates `~/.config/linearmouse/` — the one directory that setup exists to avoid. This honors the variable that defines that default.

## Behavior

Configuration is read from the first of these that exists:

| # | Path |
|---|------|
| 1 | `$XDG_CONFIG_HOME/linearmouse/linearmouse.json` — only when `XDG_CONFIG_HOME` is an absolute path |
| 2 | `~/Library/Application Support/linearmouse/linearmouse.json` |
| 3 | `~/.config/linearmouse/linearmouse.json` |

When nothing exists yet, the file is created at (1) if `XDG_CONFIG_HOME` is set, otherwise at (3) as before. Creating it at `~/.config` regardless would leave exactly the stray directory this is meant to prevent.

XDG is placed first because setting the variable is an explicit, opt-in statement of intent, and should outrank an implicit default. Happy to move it below Application Support if you'd prefer the smaller blast radius.

**Unset `XDG_CONFIG_HOME` means byte-for-byte identical behavior to today.** Per the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/latest/), values that are unset, empty, or not absolute are ignored — so a stray `XDG_CONFIG_HOME=config` in someone's environment cannot redirect anything.

## Note for anyone testing this

LinearMouse is an `LSUIElement` agent launched by Finder/`launchd`, never by a login shell, so it does **not** see `XDG_CONFIG_HOME` exported from `~/.zshrc`. It has to be set for the login session:

```sh
launchctl setenv XDG_CONFIG_HOME "$XDG_CONFIG_HOME"
```

then restart LinearMouse. Launching the app from a terminal would inherit the shell environment and prove nothing. This caveat is documented in `Documentation/Configuration.md`.

## Changes

- `ProcessEnvironment.xdgConfigHome` — the env read, split into a pure `xdgConfigHome(from:)` over an injected dictionary plus a thin computed property, so it is unit-testable.
- `ConfigurationState.configurationPaths` — prepends the XDG candidate when valid.
- `ConfigurationState.defaultConfigurationPath` — new, names the write target explicitly. This replaces the `configurationPaths.last!` force-unwrap, which was only safe because the `~/.config` append was unconditional.
- `Documentation/Configuration.md` — documents the full candidate list. It previously claimed only `~/.config`, omitting the Application Support path added in #651.

## Testing

Unit tests added to `ProcessEnvironmentTests` cover absolute / unset / empty / relative / `~`-prefixed / non-standardized values.

`ConfigurationState.configurationPaths` itself is not covered — it reads the real process environment and real home directory with no injection seam, and adding one seemed out of proportion to this change. Happy to add it if you'd rather.


lujjjh:
Thanks. When `~/Library/Application Support` was added in #651, I intended to eventually move the configuration there, keeping `~/.config` only for backward compatibility. Since XDG is not a macOS standard, I’m not sure LinearMouse should support `XDG_CONFIG_HOME`. Could you share more about the use case?

Mittelwelt:
About my use case I'll speak about;

I keep my whole config tree outside `$HOME`.

`XDG_CONFIG_HOME` points at `~/Library/0.UserLibrary/config` in macOS, which is a git repo shared between my macOS and Linux machines.

Everything else I run on macOS honors it, this change would let me remove also ~/.config, without making me maintain another directory for configs, improves portability.

So the ask isn't really "adopt XDG as a platform convention". It's "don't hardcode the default location when the user has explicitly said where their config lives".

On the "Since XDG is not a macOS standard" side, that's right but also already "~/.config" is not a macOS standard and project still uses it. If so why not also $XDG_CACHE_HOME?

There's something more: since the program is launched via launchd, so that it does not inherit XDG and any other environment variables by default. On my macOS system, I use a window manager and set it to launch at login (from system settings) and handle the launchctl thing from there with this kind of usage:

```toml
after-startup-command = [
	'''exec-and-forget /bin/zsh -c '
	## Set global environment variables for all apps.
	launchctl setenv XDG_DATA_HOME "$XDG_DATA_HOME"
	launchctl setenv XDG_STATE_HOME "$XDG_STATE_HOME"
	launchctl setenv XDG_CACHE_HOME "$XDG_CACHE_HOME"
	launchctl setenv XDG_CONFIG_HOME "$XDG_CONFIG_HOME"

	## Other
	....
	'
	''',
	'exec-and-forget open -a "..."'
]
```

This way I'm able to use the XDG variables for also GUI apps.
Same thing can be done also with a script that launches at start, that is added to Login Items, by user.

But of course, this is user's job so.

lujjjh:
Thanks for explaining. I understand the use case, and it makes sense.

I looked at a selection of popular open-source macOS apps, and most of them do not honor `XDG_CONFIG_HOME`. Karabiner-Elements is a notable exception that supports `XDG_CONFIG_HOME`. However, [its documentation](https://karabiner-elements.pqrs.org/docs/help/advanced-topics/set-environment-variables/#:~:text=Unless,link,-%2E) recommends using a directory symbolic link for relocating the configuration unless there is a strong reason to use XDG environment variables.

In hindsight, using `~/.config` was the wrong choice for a native macOS app. LinearMouse already prefers an existing configuration under `~/Library/Application Support`, while the `~/.config` path is kept only to avoid breaking existing installations. I intend to complete the migration to Application Support and deprecate the legacy path in the next major version.

Adding `XDG_CONFIG_HOME` now would introduce another supported configuration location and a new long-term compatibility contract, while the goal is to consolidate configuration under Application Support. For your setup, placing the existing configuration at `$XDG_CONFIG_HOME/linearmouse/linearmouse.json` and symlinking `~/Library/Application Support/linearmouse` to `$XDG_CONFIG_HOME/linearmouse` should provide the same portability.

Mittelwelt:
Fair enough and thanks anyways for your interest.

But IDK if it's about Homebrew version or known by you or something, but the thing is Linearmouse wrote it's config from start to NOT to ~/Library/.. but TO ~/.config on my installation. Actually that was the whole reason of emerge of this PR. Adding this as a note here.

Thanks again and have a great day.

PR #1381, opened 2026-09-04
Only observe drag events when a configured feature needs them

### LINKED ISSUE #1380: Click-and-drag stutters with an 8000 Hz mouse: the synchronous event tap observes every drag event even when no feature needs them
### OS

macOS 15.3.1 (Build 24D70)

### LinearMouse

v0.11.4

### Describe the bug

With a high polling-rate mouse (Keychron M6 8K, wired, 8000 Hz — `ReportInterval = 125` µs in `ioreg`), **click-and-drag stutters badly while LinearMouse is running**: the window/selection lags behind the physical mouse and the animation frame rate drops. Plain pointer movement is smooth; only dragging is affected.

- Quit LinearMouse → dragging is instantly smooth again.
- Relaunch LinearMouse (fresh process, minutes old) → stutter is back, 100% reproducible.
- Set the mouse to 1000 Hz → smooth even with LinearMouse running.

It also reproduces on a machine that is otherwise idle (M4 Max, GPU idle, no thermal pressure), so it is not a load issue.

#### Root cause

`GlobalEventTap` observes `EventType.all`, which unconditionally includes `leftMouseDragged`, `rightMouseDragged` and `otherMouseDragged` (but not `mouseMoved`, which is exactly why only *dragging* is affected). The tap is created with `.defaultTap` at `.cghidEventTap`, i.e. it is synchronous: WindowServer has to round-trip every observed event through LinearMouse before it can deliver it. At 8000 Hz that is up to 8000 blocking round trips per second while a window is being dragged.

Evidence:

- `sudo sample WindowServer` during a drag shows SkyLight `add_event_vector_to_tap` as the hottest frame.
- A small `CGGetEventTapList` dump shows LinearMouse's tap is the only *enabled synchronous* tap on the system whose mask contains `leftMouseDragged`; the other taps are listen-only or keyboard/scroll only. Its recorded latency spikes into the hundreds of milliseconds.
- In my configuration nothing can consume drag events for this device — the scheme only sets `pointer.disableAcceleration` and `scrolling.modifiers`. Drag events are only useful to button mappings, `switchPrimaryButtonAndSecondaryButtons`, `universalBackForward`, `autoScroll` and `gesture`.

Even at 1000 Hz this is avoidable overhead (see #915); at 4000–8000 Hz it is the difference between usable and not. #1166 had the same symptom but a different cause (fixed in 0.11.1).

### To reproduce

1. Use a mouse at 4000/8000 Hz (e.g. Keychron M6 8K wired at 8000 Hz). Check with `ioreg -r -c IOHIDDevice -l | grep -A3 'M6 8K' | grep ReportInterval` → `125`.
2. Configure LinearMouse with only pointer/scrolling settings for that device — no button mappings, no auto scroll, no gesture:

   ```json
   {
     "if": { "device": { "category": "mouse", "productID": "0xd049", "vendorID": "0x3434" } },
     "pointer": { "disableAcceleration": true },
     "scrolling": { "modifiers": { "vertical": { "command": { "type": "changeSpeed", "scale": 3 } } } }
   }
   ```
3. Drag any window around → stutters.
4. Quit LinearMouse → smooth. Relaunch → stutters. Switch the mouse to 1000 Hz → smooth.

### Expected behavior

Dragging should stay smooth when no configured feature needs drag events.

Proposed fix: derive the tap's event mask from the configuration instead of always observing all drag events —

- only include `*MouseDragged` for buttons that some scheme can actually observe while held: trigger buttons of `buttons.mappings` (including `simultaneous` / `whileHeld`), left+right when `switchPrimaryButtonAndSecondaryButtons` is on, back/forward for `universalBackForward`, and the `autoScroll` / `gesture` trigger buttons;
- keep the existing rule for `mouseMoved`;
- recreate the tap when the required set changes (this also means enabling `autoScroll` / `gesture` / `redirectsToScroll` starts observing `mouseMoved` without restarting the app).

I'm testing a patch along these lines locally and will follow up with results / a PR once verified.

### Anything else?

Hardware: MacBook Pro 16" M4 Max, two 120 Hz displays, Keychron M6 8K (USB `0x3434:0xd049`) in wired 8000 Hz mode.

eret9616:
Fixes #1380

## Problem

`GlobalEventTap` observes `EventType.all`, which unconditionally includes `leftMouseDragged`, `rightMouseDragged` and `otherMouseDragged`. The tap is a synchronous `.defaultTap` at the HID level, so WindowServer has to round-trip every one of those events through LinearMouse before delivering it. With a 4000–8000 Hz mouse that is up to 8000 blocking round trips per second while a window is being dragged, and dragging stutters visibly (plain pointer movement is fine because `mouseMoved` is not observed). Quitting LinearMouse makes it smooth again instantly.

In many configurations nothing can consume drag events at all — only pointer/scrolling settings are in use — so observing them is pure overhead.

## Change

- `EventType.required(for: schemes)` derives the tap's event mask from the configuration:
  - the base set (scroll wheel, button down/up, key down/up, flags changed) is always observed;
  - `*MouseDragged` is only observed for buttons that some scheme can actually watch while held: trigger buttons of `buttons.mappings` (including `simultaneous` / `whileHeld`), left + right when `switchPrimaryButtonAndSecondaryButtons` is on, back/forward for `universalBackForward`, and the `autoScroll` / `gesture` trigger buttons (defaulting to the middle button, matching the transformers);
  - `mouseMoved` follows the existing rule (`redirectsToScroll`, `autoScroll`, `gesture`).
- `GlobalEventTap` subscribes to `ConfigurationState.$configuration` and recreates the tap when the required set changes (debounced, on the main run loop). As a side effect, enabling `autoScroll` / `gesture` / `redirectsToScroll` now starts observing `mouseMoved` without restarting the app.
- The observed event types are logged at info level when the tap is created, which makes this easy to check with `log show`.

The mask is still global (one tap), so a mapping on button 3 for one device keeps `otherMouseDragged` observed for all devices — exactly what happens today for those buttons; the change only ever removes drag types that no feature could consume.

## Testing

- `xcodebuild test`: 841 tests, 0 failures.
- Manual, macOS 15.3.1, Keychron M6 8K wired at 8000 Hz, config with only pointer/scrolling settings for that device plus button mappings (buttons 2/3) on other devices:
  - stock v0.11.4: window dragging stutters 100% of the time; this branch: smooth.
  - `CGGetEventTapList` shows the tap mask going from `0xe401cde` to `0xe401c1e` (`leftMouseDragged` / `rightMouseDragged` dropped, `otherMouseDragged` kept because of the button 2/3 mappings).
  - Scrolling, the Command scroll-speed modifier and the button mappings keep working.
- Note: `main` currently needs the Xcode 26 SDK because of `glassEffect` in `SecondaryButtonStyle.swift`; I built and ran the tests with Xcode 16.4 by stubbing that single line locally (not part of this PR).


lujjjh:
Thanks for investigating this. The optimization makes sense as a workaround, although the performance issue still remains when these features are actually in use.

There is also a correctness issue in `EventType.required(for:)`. It examines each raw scheme independently, while Auto Scroll settings are merged field by field. For example, if a base scheme enables Auto Scroll with the middle button and a conditional scheme overrides only the trigger to the right button, the effective configuration uses the right button. However, the conditional scheme has no explicit `enabled: true`, so it is skipped and `rightMouseDragged` is omitted from the event mask.

More generally, `EventType` and `GlobalEventTap` should not need to duplicate the configuration rules of every feature. It may be cleaner for transformers to declare their required event types through the same path used to build them, with the global event tap observing the resulting union across the configuration.

Issue #757, opened 2024-07-01
Settings for Magic mouse to mimic default MacOS trackpad gestures


baleeds:
### OS

macOS 14.5 (Build 23F79)

### LinearMouse

v0.10.0

### Is your feature request related to a problem?

I'm frustrated that the default gestures on the magic mouse don't match the trackpad exactly. For instance, there's no way to trigger the Expose feature. Mission control uses taps instead of a swipe.

### Describe the solution you'd like

I would like to have settings that alter the gestures so they more closely match the trackpad, except with two fingers instead of three. I do understand that swiping up and down with two fingers will change the default behavior which is to scroll even with one finger resting on the surface.

My desired gestures would be:
- Two-finger swipe left/right to navigate workspaces
- Two-finger swipe up for Mission Control
- Two-finger swipe down for Expose

The settings could be in a page called "Gestures". There would be the following options:
- Two-finger double tap (Default behavior, No action, Mission Control, Expose, Launchpad, Show Desktop)
- Two-finger swipe left/right (Default behavior, No action)
- Two-finger swipe down (Default behavior, Expose, Mission Control, Launchpad, Show Desktop)
- Two-finger swipe up (Default behavior, No action, Expose, Mission Control, Launchpad, Show Desktop)

In these scenarios, "Default Behavior" would not capture the event at all, but would leave the event to be dealt with by the OS. "No action" would intercept the event and do nothing.

### Describe alternatives you've considered

I've attempted to use BetterTouchTool to create my own gestures, but they don't feel optimized in a way that's indicative of a native feature.
I could get over it and move on with my life, but it feels inconsistent, and like the touch surface on the mouse is underused.

PR #1382, opened 2026-09-07
Require two fingers to scroll on Magic Mouse

corinna000:
## What

  Adds an opt-in **"Require two fingers to scroll"** toggle (Scrolling settings, only shown when a Magic Mouse is the selected device) that
  requires two fingers on the touch surface before scrolling starts, matching trackpad behavior instead of the Magic Mouse's default of scrolling
  on any single-finger contact.

  Off by default - existing behavior is unchanged unless a user opts in.

  ## Why

  The Magic Mouse scrolling on any single touch (rather than requiring two fingers, like a trackpad) causes frequent accidental scrolling from
  incidental palm/finger contact. This has come up before with no resolution:

  - #757 - feature request for trackpad-like Magic Mouse gestures, closed stale
  - Discussion #758 - this exact request ("Disable single finger scroll on Magic Mouse but keep 2 finger scroll"), never answered

  ## How

  There's no public API for reading finger-count data off a Magic Mouse. This uses Apple's private, undocumented `MultitouchSupport.framework` -
  the same one BetterTouchTool, Mos, and the (now largely unmaintained) MagicPrefs have historically relied on for the same purpose. I want to
  flag this clearly up front since I know `IOHIDManager`/public APIs are preferred where possible (per the discussion on #200) - I don't believe
  a public alternative exists for multitouch contact data on this device, but happy to be corrected.

  Given the private-API dependency, this is designed to fail safe everywhere:
  - `MultitouchSupportBridge` resolves every symbol via `dlopen`/`dlsym` and reports `isAvailable = false` if anything is missing (e.g. on a
  future macOS release), rather than crashing.
  - `RequireTwoFingerScrollTransformer` fails open (never blocks scrolling) whenever the finger-count signal isn't available.

  One correctness detail worth calling out for review: gating is based on live finger count for **every** scroll event, deliberately not on
  `CGMomentumScrollPhase`. A trackpad reserves that field for the post-lift inertial tail, but the Magic Mouse has no real gesture-phase concept
  - empirically it tags most actively-touched scroll ticks (not just the glide after lift-off) as "momentum changed". Trusting that field let
  most one-finger scrolling through unfiltered during development. Instead, the transformer tracks its own "was this gesture qualified with two
  fingers" state (with a time-gap fallback for when the device never sends an explicit momentum-end), so a real two-finger scroll still glides to
  a natural stop after lift-off without that grace period leaking into unrelated one-finger scrolling later.

  ## Testing
  
  - New unit tests for `RequireTwoFingerScrollTransformer` covering suppression, pass-through, the fail-open path, synthetic-event bypass, and
  the gesture-qualification/momentum-glide logic (including the regression this exists to prevent - trusting `momentumPhase` alone for a fresh,
  unqualified gesture).
  - Full existing suite still green.
  - Manually verified against a physical Magic Mouse: one-finger scroll blocked, two-finger scroll works, two-finger scroll glides naturally
  after lifting fingers mid-gesture.
  
  ## Config

  New `scrolling.requireTwoFingerScroll` boolean field (documented in `Configuration.d.ts`, regenerated JSON schema), off by default.


corinna000:
@lujjjh What do I need to do to push this forward? Thank you!

lujjjh:
Thanks for working on this!

At first glance, `firstExternalDevice()` may have an issue since an external Magic Trackpad could also be picked up there. If all goes well, I’ll test this on a Magic Mouse and do a more detailed review tomorrow.

I’m also thinking about how this should be modeled in LinearMouse. `requireTwoFingerScroll` feels a bit too specific. Maybe Magic Mouse gestures like one finger and two fingers should be treated as inputs that can be mapped to different actions or behaviors. I’d like to think a bit more about the scope here.

lujjjh:
I tested this on a physical Magic Mouse. On my end, two-finger scrolling is inconsistent: it often fails to start, or loses momentum. Once scrolling starts, lifting one finger can still work normally, so it does not seem to be simply related to releasing both fingers together.

I also haven’t settled on how this should be modeled in the configuration or what its shape should be yet. We can continue discussing that.

corinna000:
Thank you for the comments! 
1. I don't have an external Magic Trackpad for testing, but your comment makes a lot of sense. I'll see if there's a way to gate the code so it's only on the Magic Mouse. 
2. The surface of the Magic Mouse is multi-touch, so one risk of accepting this PR is that if you wanted to support pinch/zoom functionality in the future then change may require a refactor. 
3. The Magic Mouse I'm testing on is the old, battery driven model. It's entirely possible the newer models have different behaviors, which if true would be sad. Short of my finding a newer model I'm not sure how to move forward. But I can partially confirm your observation by adding this clarification: Once a two-finger scroll is started it no longer counts the number of mice on the device, so you can lift your second finger and still scroll until the first finger (so zero contact with the mouse) remains. This may explain some of the jank.

PR #1392, opened 2026-09-12
Allow swipe recording after a provisional long press

### LINKED ISSUE #1391: Button Mappings recorder cannot record swipes after provisional long press
### Version

v0.12.0-beta.5

### Problem

While configuring an MX Master 3 thumb button (Logitech Control 0x00C3), the advanced Button Mappings recorder repeatedly recognizes Long Press instead of a swipe. Adding the desired short-press and four swipe outcomes directly to the configuration works, so these actions can be configured without the recorder. Independence from Options+ is being checked separately; this report does not claim a confirmed HID movement-delivery cause.

There is a deterministic recorder problem in `ButtonMappingRecordingEngine`: after 500 ms, `advance(to:)` sets recognition to `.longPress`, and `pointerMoved` rejects further movement because it requires `recognition == nil`. A user who pauses before dragging cannot record a swipe. Ordered button and wheel inputs already replace this provisional long press while the button remains held.

### Reproduction in the recording engine

1. Send button down for Logitech Control 0xC3 at 0 ms.
2. Advance to 500 ms: the preview becomes Long Press.
3. Send two straight upward movement deltas of 30 points each at 700 and 800 ms.
4. Release at 900 ms.

Actual: Long Press remains recorded.
Expected: Swipe Up replaces the provisional Long Press once the swipe threshold is reached. Long Press should remain selected when released without sufficient movement.

### Proposed scope

Allow movement to replace a provisional long press in the recorder, matching its existing ordered-button and wheel recording behavior. Preserve completed outcomes and the runtime action recognizer's commitment rules.

lamhoangtung:
When a user holds a button for 500 ms before dragging in the advanced Button Mappings recorder, the preview becomes Long Press and all later movement is ignored. This makes recording a swipe require racing the long-press timer.

Allow a swipe to replace the provisional long press while the button remains held, matching the recorder's existing behavior for ordered button and wheel inputs. Release without sufficient movement still records Long Press; committed swipe and released outcomes remain unchanged. Runtime recognition is unchanged.

Closes #1391.

Validation: 21 ButtonMappingRecordingEngineTests passed with `xcodebuild test` on macOS. Regression tests cover all four directions with Logitech Control 0xC3, accumulated movement after the long-press threshold, and movement after release. SwiftFormat and SwiftLint passed for the changed files.


LinearMouse/UI/ButtonsSettings/ButtonMappingsSection/ButtonMappingRecordingEngine.swift
```diff
@@ -138,9 +138,12 @@ struct ButtonMappingRecordingEngine {
 
     mutating func pointerMoved(deltaX: Double, deltaY: Double, at timestamp: UInt64) {
         advance(to: timestamp)
+        // Long press is provisional until release, just as it is for ordered
+        // buttons and wheel inputs. Keep collecting movement while held so a
+        // user can record a swipe without racing the long-press timer.
         guard !pressedButtons.isEmpty,
               wheelMapping == nil,
-              recognition == nil else {
+              recognition == nil || recognition == .longPress else {
```
lujjjh, line 146: This also allows a long press to become a swipe after the trigger button has already been released, as long as another button remains held.

For example:
1. Hold A, then press and hold B until Long Press is recognized.
2. Release B while keeping A held.
3. Move the pointer past the swipe threshold.

The recorder now produces “hold A + swipe B,” even though B was already released before the movement.

Could we require all buttons participating in the recorded trigger to remain held before allowing this reclassification, or freeze the outcome on the first release? The new release test only covers a single button; ordered and simultaneous combinations with both release orders would help cover this case.

Issue #1378, opened 2026-09-02
Previous space trigger not working


jimjamjar021:
### OS

macOS 27.0 (Build 26A5425a)

### LinearMouse

v0.11.4

### Describe the bug

when using the triggers like launchpad and mission control in linear mouse, the next space option works perfectly but the previous space option just doesn't do anything even if I do the trigger for it, I tried changing the button to activate the trigger like the primary and secondary button and every time I retried it, the next space action worked perfectly but the previous space action did nothing.

### Expected behavior

At the activation of the trigger the space must move to the one to the left. For example in two spaces desktop 1 being the first and desktop 2 being the second, the space must move from desktop 2 to desktop 1.

Yash121l:
I could not find any asymmetry between the two space actions in LinearMouse itself, but there is one in the system lookup they both use: `CGSGetSymbolicHotKeyValue` returns success with a virtual key of `0xFFFF` (or `0`) when the Mission Control shortcut has been cleared or never assigned, and LinearMouse posted that as a key event, so the action failed silently or sent a stray ⌃A. #1393 makes that case throw and log `symbolicHotKeyNotConfigured` instead. If that is what you are hitting, check System Settings > Keyboard > Keyboard Shortcuts > Mission Control and make sure "Move left a space" still has ⌃← assigned.

PR #1393, opened 2026-09-13
Do not post a stray key when a symbolic hot key has no key assigned

Yash121l:
## What broke

In #1378 "Mission Control: move right a space" works while "move left a space" does nothing. The dispatch for the two actions is symmetric (`ButtonActionExecutor.swift:507-511`, `GestureButtonTransformer.swift:285-289`, `SymbolicHotKey.spaceLeft = 79` / `spaceRight = 81` matching `CGSHotKeys.h`), so the only way the two can diverge is the system hot-key lookup.

## Why

`CGSGetSymbolicHotKeyValue` returns `success` even when the hot key has no key assigned. I probed it on macOS 26.6.2:

```
hk=79  err=0 keyEquiv=0xffff vk=123   mods=0x840000 enabled=1   (⌃←, assigned)
hk=83  err=0 keyEquiv=0xffff vk=0     mods=0x40000  enabled=0   (never assigned)
hk=164 err=0 keyEquiv=0xffff vk=65535 mods=0x0      enabled=0   (cleared in System Settings)
```

`postSymbolicHotKey` posted those values as they came: virtual key `0xFFFF` does nothing, and `0` is the A key, so the action could send ⌃A to the frontmost app. Either way it failed silently, with no log line. If a user cleared or lost "Move left a space" while "Move right a space" is untouched, that is exactly what #1378 looks like. I also confirmed that `CGSSetSymbolicHotKeyEnabled(hotKey, true)` does not restore a value for a cleared hot key, so the existing enable/disable path cannot recover it.

## What changed

`postSymbolicHotKey` now rejects a value with no usable virtual key code and throws `CGSError.symbolicHotKeyNotConfigured` before touching the hot key's enabled state. `ButtonActionExecutor` already logs thrown errors, so the failure is visible instead of silent, and no stray keystroke is sent.

I did not add a fallback that posts the default ⌃← / ⌃→ combination: a cleared symbolic hot key does not respond to its default combination, so that would only look like a fix. If you would rather have LinearMouse temporarily assign the default value via `CGSSetSymbolicHotKeyValue` and restore it afterwards, I will do that in a follow-up.

## How it was verified

- New `KeyKitTests.testSymbolicHotKeyVirtualKeyCodeRejectsHotKeysWithoutAKey` covers the cleared (`0xFFFF`), never-assigned (`0` with no key equivalent), normal arrow-key, and genuine A-key cases.
- `swift test --package-path Modules/KeyKit`: 13 tests, 0 failures.
- `xcodebuild test -project LinearMouse.xcodeproj -scheme LinearMouse -destination 'platform=macOS'`: 920 tests, 0 failures.
- `swiftformat --lint .` clean; `swiftlint .` reports no new violations.

For the reporter of #1378: after this change, `log stream --predicate 'subsystem == "com.lujjjh.LinearMouse"'` shows `Failed to execute: missionControl.spaceLeft: symbolicHotKeyNotConfigured` if the cause is a missing system shortcut. The workaround is to re-assign ⌃← under System Settings > Keyboard > Keyboard Shortcuts > Mission Control > "Move left a space".

Refs #1378


lujjjh:
Thanks!

PR #1395, opened 2026-09-14
Filter configuration watcher events without resolving symlinks per event

### LINKED ISSUE #1368: High CPU usage in LinearMouse 0.11.4 due to broad FSEvents configuration watches
## Description

LinearMouse 0.11.4 consumes 18% to 51% of a CPU core on macOS 26.6.1 while the mouse event thread is idle. This load recurs within 10 seconds of relaunching the application, indicating it is not a stale event backlog.

Active configuration file:
`~/.config/LinearMouse/linearmouse.json`

The FSEvents stream watches both `~/.config` and `~/Library/Application Support`. Unrelated writes under these roots force LinearMouse to canonicalize every event path on the main thread before filtering.

## Sample Evidence

An 8-second sample of the original 0.11.4 process captured 2,110 of 6,863 main-thread samples (about 31%) in the FSEvents callback.

A second sample from a fresh 0.11.4 process captured 2,013 of 8,740 main-thread samples (about 23%) in the callback, including 332 `readlink` and 73 `lstat` samples. The mouse event thread remained idle in both traces.

PR #1209 introduced the broad watcher. In `FileWatcher.rootPaths(for:)`, the watch root for a configuration file defaults to its parent directory. For `~/.config/LinearMouse/linearmouse.json`, this expands the watch to the entirety of `~/.config`. Additionally, `ConfigurationState.startHotReload()` supplies both candidate paths, adding all of `~/Library/Application Support` when the alternate configuration directory is missing.

## Proposed Fix and Validation

The proposed patch resolves this by:

1. Supplying only the active `configurationPath` to `FileWatcher`.
2. Watching the existing configuration directory directly instead of its parent.
3. Falling back to the nearest existing ancestor only when the configuration directory is missing, then rebinding upon recreation.

The focused `FileWatcherTests` suite passed all tests (including directory deletion and recreation), and the full Xcode test run completed with all 561 test cases passing.

In an 8-second sample of the patched build with Accessibility enabled, the main thread was idle for 6,824 of 6,879 samples. There were no `FileWatcher`, FSEvents, `readlink`, or `lstat` frames. The signed 0.11.2 release remains at 0.0% CPU as a practical workaround.

The tested 3-file patch is ready to submit.

## Environment

- LinearMouse 0.11.4
- macOS 26.6.1 (25G76)
- Apple silicon
- Active configuration: `~/.config/LinearMouse/linearmouse.json`

anandghegde:
Credit to @mhaslinsky for the diagnosis in #1368.

## Cause

`FileWatcher` watches the parent of each configuration directory, so with `~/.config/linearmouse/linearmouse.json` it receives every event under `~/.config` (and under `~/Library/Application Support` for the other candidate path). For each callback it canonicalized every event path (readlink/lstat per path component) on the main queue, and then rebuilt the watch roots, before filtering the events out.

## Fix

The watch roots and the candidate paths stay as they are, so deleting the higher-priority file and retargeting a symlinked configuration directory keep working. Only the filtering changes:

- FSEvents reports paths with their directories already resolved; only a symlink in the last component differs from its canonical form. The paths that matter (the file and its directory with the last symlink preserved, their canonical forms, and the exact parent) are resolved once when the stream is updated, and event paths are compared against them as reported.
- The stream is refreshed only after an event that touches the watched files or their ancestors, because only those can move the roots.

## Testing

- New `FileWatcherTests`: sibling-directory events are not relevant; a symlinked configuration directory is; deleting the higher-priority file is reported; retargeting a symlinked directory is reported, and so is a later edit in the new target.
- `xcodebuild test -project LinearMouse.xcodeproj -scheme LinearMouse -destination 'platform=macOS'`: 924 tests, 0 failures. `FileWatcherTests` with `-test-iterations 20`: all passed.
- `swiftformat --lint` and `swiftlint lint` on the changed files: clean.
- Command-line harness compiled with `FileWatcher.swift`, watching `<tmp>/a/linearmouse/linearmouse.json` and `<tmp>/b/linearmouse/linearmouse.json` while a script rewrites files in a sibling `other-app` directory in a tight loop. CPU time from `getrusage` over 15 s:

  | | main | this PR |
  |---|---|---|
  | writer in `a` | 58.5% of a core | 0.3% |
  | writers in `a` and `b` (`b/linearmouse` missing) | 106.3% | 0.5% |

- Same harness, both regressions from review: deleting the higher-priority file fires a change callback, and retargeting a symlinked `linearmouse` directory fires one for the retarget and another for a later edit in the new target. Both match `main`.

Fixes #1368

https://claude.ai/code/session_01MA8JGs7pCoXFL2W6jEJ8sD


lujjjh:
Thanks! It seems that there are two regressions:

- When both config files exist, deleting the higher-priority file no longer reloads the fallback config.
- When the config directory is a symlink, pointing it to a different directory can stop hot reload, even for later edits in the new directory.

anandghegde:
Thanks, you're right on both. Narrowing the watch was the wrong lever. 7dcbacf goes back to the original watch roots and both candidate paths, and fixes the CPU where it was spent instead: event paths are now compared against paths resolved once when the stream is updated (FSEvents already reports directories resolved), and the stream is only refreshed after an event that touches the configuration files.

Both cases have tests now (`testReportsDeletionOfHigherPriorityFile`, `testReportsChangesAfterSymlinkedDirectoryIsRetargeted`), and I checked them against `main` and the previous commit with a small harness: the previous commit missed both, this one reports them like `main`. Under a busy sibling directory the watcher uses 0.3–0.5% of a core versus 58–106% on `main`. I've updated the description with the details.

lujjjh:
Thanks for the update! Looks good to me.

PR #1400, opened 2026-09-19
Add a hold trigger for converting pointer movement to scroll

### LINKED ISSUE #1065: Easy way to toggle on/off 'convert pointer movement to scroll'
### OS

macOS 26.2 (Build 25C56)

### LinearMouse

v0.10.2

My problem is that I hate scroll wheels, they really bother my fingers, even the infinite scrolling type. Another problem is when Mac Finder is in columns and I need to horizontally scroll, I have no way to do this. I want a better way to scroll that doesn't hurt my finger and includes horizontal scrolling. 

I really love the 'convert pointer movement to scroll events' feature, but there's no easy way to turn it on and off. I've tried using other tools to execute scripts to update the linearmouse settings file, but the app doesn't seem to register the change. I've actually added a second mouse to my desk and configured that one to scroll with movement, but I don't like it for a long term solution. What I think would be ideal is:
- scroll wheel works as normal when rotating
- click + hold scroll wheel mouse turns into scrolling with movement device
- release scroll wheel mouse returns to normal

Other ways it could be handled are with a modifier key, but I like to take my hand off the keyboard when I'm reading something, it helps me disengage from "input mode" and focus on the content. Mapping a mouse button to toggle the mode on/off with a click would also work. 

The most simple thing that would satisfy the request is just to add a keyboard shortcut to toggle it ON. You already have one to toggle it OFF. I can configure my mouse wheel click to execute those commands and then I'm happy. 

THANK YOU!!!

### Additional context

_No response_

anandghegde:
Closes #1065.

This follows option A from #1065, which @lujjjh agreed to: a `redirectsToScrollTrigger` field on `Scheme.Pointer` that reuses `Mapping.Trigger`, moved to a shared location.

## Summary

- Add `pointer.redirectsToScrollTrigger`. While the trigger is held, pointer movement becomes scrolling with the same deltas and fixed cursor as `redirectsToScroll` today. When it's released, the pointer moves normally again.
- The trigger's press and release are consumed, so it doesn't also click or run a button mapping on the same button. Other buttons still work while it's held.
- Without a trigger, `redirectsToScroll: true` works as before (always on). The trigger only applies when `redirectsToScroll` is true.
- Move `Scheme.Buttons.Mapping.Trigger` to `Scheme.Trigger` (`Scheme/Trigger.swift`). `Mapping.Trigger` is kept as a typealias, so existing code and the JSON format don't change. `Button`, `Modifier` and `ScrollDirection` stay under `Mapping` for now. `valid(with:)` stays in `StructuredMapping.swift` as an extension because it checks mapping outcomes.
- The new field accepts a subset of `Trigger`: a mouse button with optional modifiers. Primary without modifiers, wheel inputs, chords, `whileHeld` and Logitech controls are rejected. An invalid trigger turns redirecting off rather than falling back to always on. The settings page shows the same error Auto Scroll uses.
- Settings: under "Convert pointer movement to scroll events" there's now a Trigger recorder, the same `ButtonMappingButtonRecorder` Auto Scroll uses, plus a Clear button. Recording keeps the current trigger until a new one is captured.
- Hold only. It's separate from Auto Scroll's hold mode.
- Updated `Configuration.d.ts`, the generated `Configuration.json` (`npm run generate:json-schema`) and `Configuration.md`. Added two English strings.

Example:

```json
"pointer": {
  "redirectsToScroll": true,
  "redirectsToScrollTrigger": { "input": { "button": 3 }, "modifiers": ["option"] }
}
```

Open questions:

- Modifier-only triggers, such as holding Option: `Trigger` needs an input, so I left them out. They'd need a new input kind and would overlap with the scrolling modifier settings. Should they be a follow-up?
- Logitech (HID++) controls as triggers: these need the diverted-control handling Auto Scroll has. Worth a follow-up?
- Toggle mode (click on, click off): not included. Hold only for now?

## Testing

- `xcodebuild test`: all 943 unit tests pass. New tests cover decoding, encoding and merging the field, the `Mapping.Trigger` typealias, which triggers are valid, held vs released behaviour (press/release consumed, drags redirected only while held, modifiers matched at press, other buttons pass through, bypassed while recording, reset on deactivate), and where the transformer goes in the chain.
- `swiftformat --lint .` is clean and `swiftlint` reports no new violations.
- Not yet tested with a physical mouse. Opening as a draft; I'll mark it ready after that.

(I used Claude Code to help write this. I've reviewed the changes myself.)


anandghegde:
@lujjjh one thing I ran into while testing this on my own machine, which I think needs a decision before this is mergeable.

The trigger recorder only appears once "Convert pointer movement to scroll events" is already on, and that checkbox takes effect immediately. So the only way to reach the trigger is through the always-on state, and in always-on state the selected device has no usable pointer any more. I did this twice: once on my Magic Mouse and once, by accident, on the built-in trackpad, because the Pointer tab preselects the device you are currently using. Both times the only way out was to quit LinearMouse from another input device or the terminal.

That is pre-existing behaviour of `redirectsToScroll` rather than something this PR introduces, but this PR is what makes a safe configuration possible, so it seems the right place to fix the path to it.

What I would suggest, and I am happy to implement whichever you prefer:

1. Replace the checkbox with a three-way picker: Off / While holding a trigger / Always. Picking "While holding" shows the recorder and does not write `redirectsToScroll` until a trigger has actually been recorded, so movement is never captured without a way to release it. "Always" keeps today's behaviour for people who want it.
2. Keep the checkbox, but when it is turned on with no trigger set, show the recorder and leave the setting inactive until either a trigger is recorded or the user explicitly confirms always-on.

Option 1 also makes the config shape self-documenting in the UI, since `redirectsToScroll` plus `redirectsToScrollTrigger` is effectively three states already.

Either way the JSON stays as it is in this PR.

lujjjh:
@anandghegde thanks for catching this. I'm leaning toward option 1 since it makes the three states clearer.

For "Always," maybe we could add a confirmation before apply it, or automatically revert after 10 seconds unless the user confirms, so they don’t accidentally lose pointer control.

anandghegde:
Thanks — went with option 1, and pushed it in `62cc87b`.

**The picker.** "Convert pointer movement to scroll events" is now Off / While holding a trigger / Always, and the trigger recorder only appears under the middle one. The stored configuration is unchanged — still `redirectsToScroll` plus an optional `redirectsToScrollTrigger` — so no schema regeneration was needed, and an existing scheme that only sets `redirectsToScroll: true` still reads back as "Always".

That also closed a trap I'd left in: with the old checkbox, clearing the trigger while it was ticked silently promoted the setting to convert-everything. Clearing now drops back to "While holding a trigger" with conversion off until a new trigger is recorded, so it can't happen by accident.

**On "Always".** I implemented both things you suggested rather than picking one, because neither covers the case alone:

- Confirmation alone doesn't actually protect anyone. The user clicks Turn On, the pointer stops, and they're in exactly the state the dialog warned about — with no pointer to get out of it.
- Auto-revert alone applies a pointer-disabling change with no warning at all.

Together they're the shape macOS uses for a display change: you say yes, it applies, and it goes back by itself after 10 seconds unless you keep it. `Keep` is the default action and `Revert` the cancel action, so both are reachable from the keyboard while the pointer is unusable — that part matters more than the dialog does. (The `keyboardShortcut` API is macOS 11+, so there's an `if #available` pair around it, matching the existing one in this file for the revert-to-defaults button.)

The countdown timer runs in `.common` mode so it isn't paused while a menu or slider is being tracked.

**Tests.** Added `PointerRedirectsToScrollModeTests` — six cases over the mode↔configuration reading, including one that decodes `{ "pointer": { "redirectsToScroll": true } }` and asserts it still means "Always". 943 → 949 tests, 0 failures; `swiftformat --lint .` and `swiftlint` are clean.

Happy to drop the confirmation dialog and keep only the auto-revert if you'd rather have fewer clicks — the revert path doesn't depend on it.

lujjjh:
Thanks for the update! I tested the changes in this PR, and everything works well overall. One thing I noticed is that when a trigger is already set, clicking to record a new one gives no indication that recording is in progress. The new strings also need translations. Aside from those, it looks good to me. Do you think we can mark it as ready for review?

anandghegde:
Thanks for testing it! Both are fixed in `79b84b0`, and I've marked the PR ready for review.

**Recording indicator.** The recorder kept showing the old trigger because the binding threw away the cleared mapping on purpose, so the stored trigger stays in effect until a new one is recorded. The in-progress mapping is now held separately while recording. The recorder shows "Recording" (plus any modifiers held so far), the same as the Auto Scroll and gesture recorders, and the existing trigger stays active underneath. If recording ends without a new trigger, for example by clicking the recorder again, it goes back to showing the stored trigger.

**Translations.** All ten new strings are translated into the 34 locales in the catalog, reusing the existing terms for pointer, trigger and scroll events in each language (and the same formal or informal register each locale already uses). The revert countdown uses plural variations, like `%d line(s)`, so the forms are right in languages such as Polish, Russian and Arabic.

The build and all 949 tests pass, and `swiftformat --lint` and `swiftlint` are clean on the changed files.

